PR #4, opened 2018-03-04
fix webpack version

eloytoro:
Right now tests fail because the wrong version of webpack is being installed

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "nsp": "^3.1.0",
     "pre-commit": "^1.2.2",
     "standard-version": "^4.3.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "nsp": "^3.1.0",
     "pre-commit": "^1.2.2",
     "standard-version": "^4.3.0",
-    "webpack": "^3.11.0",
+    "webpack": "^4.0.0",
```
michael-ciniawsky: `4.1.0` (https://github.com/webpack/webpack/pull/6663)

sokra:
Thanks

PR #5, opened 2018-03-04
Test case for "composes" classes

eloytoro:
It seems that if you compose to a different class it will import the entire css twice.

### Input

async-1.css
```css
:local .base {
  composes: composed from './async-2.css';
  background: blue;
}
```

async-2.css
```css
:local .composed {
  background: green;
}
```

### Desired output

1.css
```css
.base {
  background: blue;
}
```

2.css
```css
.composed {
  background: green;
}
```

### Current output

1.css
```css
.composed {
  background: green;
}

.base {
  background: blue;
}
```

2.css
```css
.composed {
  background: green;
}
```

### DIFF
--- a/test/cases/composes-async/async-1.css
+++ b/test/cases/composes-async/async-1.css
@@ -1,0 +1,4 @@
+:local .base {
+  composes: composed from './async-2.css';
+  background: blue;
+}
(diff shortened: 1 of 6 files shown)


sokra:
Not sure about that. When chunk 1 is loaded it's missing the `.composed` styling in your desired output.

Issue #2, opened 2018-03-04
publicPath doesn't seem to be honoured when loading CSS


insin:
`mini-css-extract-plugin@0.1.0`

Using an app with a router, which splits code per route and has `publicPath: '/'` in its output options, I'm seeing the following inserted into `<head>` when visiting a route which imports CSS:

![firefox_2018-03-04_22-00-22](https://user-images.githubusercontent.com/226692/36945311-b843f620-1ff7-11e8-853e-0136b73fb753.png)

If I log out `outputOptions` in `loader.js` it looks ok, but don't know what's ultimately responsible for creating the full path in the `compilation.createChildCompiler` or `NodeTemplatePlugin` it gets passed to:

```js
{ outputOptions: { filename: '*', publicPath: '/' } }
```

Kovensky:
It seems that the plugin's output doesn't take into account the `__webpack_public_path__`. 

Compare this line in the extracted CSS output:

```js
linkTag.href = "" + ({"id":"name"}[chunkId]||chunkId) + ".css";
```

with:

```js
script.src = __webpack_require__.p + "" + ({"id":"name"}[chunkId]||chunkId) + .js";
```

For `[name].js` / `[name].css` outputs.

ganapativs:
Same issue here.

PR #7, opened 2018-03-05
fix(index): Use the runtime publicPath when generating the href

Kovensky:
Right now all generated URLs end up being top level URLs.

Fixes https://github.com/webpack-contrib/mini-css-extract-plugin/issues/2

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -172,7 +172,7 @@ class MiniCssExtractPlugin {
                   'linkTag.rel = "stylesheet";',
                   'linkTag.onload = resolve;',
                   'linkTag.onerror = reject;',
-                  `linkTag.href = ${linkHrefPath};`,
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -172,7 +172,7 @@ class MiniCssExtractPlugin {
                   'linkTag.rel = "stylesheet";',
                   'linkTag.onload = resolve;',
                   'linkTag.onerror = reject;',
-                  `linkTag.href = ${linkHrefPath};`,
+                  `linkTag.href = ${mainTemplate.requireFn}.p + ${linkHrefPath};`,
```
Kovensky: This was just copied from the JsonpMainTemplatePlugin.

sokra:
Thanks

PR #9, opened 2018-03-05
fix(loader): Correctly preserve CSS module locals

Kovensky:
The handling was there, but the property it tried to access was being clobbered by the .map's result.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -87,9 +87,10 @@ export function pitch(request) {
     if (!source) {
       return callback(new Error("Didn't get a result from child compiler"));
     }
-    let text;
+    let text, locals;
     try {
       text = exec(this, source, request);
+      locals = text && text.locals;
       if (!Array.isArray(text)) {
         text = [[null, text]];
       } else {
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

Onelin:
**Css-loader is packed, but the div's class does not take effect.
Using extract-text-webpack-plugin@4.0.0beta is in effect.
help me look, thanks**

image
image
image

sylv3r:
Same issue, the finale css file is populated with classes, as expected, but my `class` attributes are empty everywhere I used one of this class. This isn't fixed.

I went back to use extract-text-webpack-plugin@4.0.0beta which works as expected.

evilebottnawi:
@Onelin @sylv3r can you create minimum reproducible test repo? Also please create new issue. Thanks!

sylv3r:
@evilebottnawi @Onelin I actually found the solution while trying to prepare the MRE..

One needs to switch between `style-loader` and `MiniCssExtractPlugin.loader` as the first loader use by the rule depending on the env (ie. `development` for HMR or `production` for dist builds) since this plugin is not HMR ready, yet.

**Short example (without loaders options) :**
```js
{
  test: /\.s?[ac]ss$/,
  use: [
    NODE_ENV === 'production' ? MiniCssExtractPlugin.loader : 'style-loader',
    'css-loader',
    'postcss-loader',
    'sass-loader',
  ],
}
```

I think this should be explain in the docs since this is a bit tricky.

evilebottnawi:
@sylv3r Can you do PR?

sylv3r:
Sure, have a look at #101 @evilebottnawi

Onelin:
I solved my problem by using the MiniCssExtractPlugin in the use:[], but not using the 'style-loader' in front of it, causing a conflict. Tks

Issue #1, opened 2018-03-04
Content hash support


insin:
Using `mini-css-extract-plugin@0.1.0`

Is it possible to support a content hash in the filename, like extract-text-webpack-plugin's `[contenthash]`?

Using `filename: '[name].[chunkhash:8].css'` the `.css` hash is also changing when only the JavaScript in the chunk has changed:

![conemu64_2018-03-04_21-35-20](https://user-images.githubusercontent.com/226692/36945132-01f2e4aa-1ff5-11e8-9b0c-20488d2e2180.png)

![conemu64_2018-03-04_21-42-07](https://user-images.githubusercontent.com/226692/36945134-0498c95e-1ff5-11e8-9d1f-c7ac101f5b36.png)

developer-lindner:
Needs to be added! There were some issues with Cache Busting not working. 
contentHash helps there!

evilebottnawi:
FYI. From slack:
```
sokra [3:19 PM]
I'll take care of `[contenthash]` this week. It will be added to webpack core, because it is also needed for JS files in combination with CSS or WASM files.
```

It is not necessary to implement this here :star: Just wait next release `webpack@4`

marcusdarmstrong:
Shouldn't the chunk hash include the contents of the CSS sources, just as it does for the JS sources?

Kovensky:
That's what the chunk hash does, yes. This is the content hash for the individual asset.

marcusdarmstrong:
Yes; but the CSSModule defined by this plugin doesn't update the chunk hash—it just relies on the updateHash implementation in Module.

Perhaps that should be treated as a separate issue?

sokra:
The impl has finished. CSS module test has also been added. Just waiting for webpack 4.3.0 release with contenthash support to send a PR.

philipstanislaus:
@sokra thanks, this is awesome! Webpack 4.3.0 has just been released.

bebraw:
This should work now.

bebraw:
Sorry, my mistake. #59 has to land first.

PR #30, opened 2018-03-09
feat(index): add `[contenthash]` support

chenxsan:
### `Issues`

- Fixes https://github.com/webpack-contrib/mini-css-extract-plugin/issues/1

First time doing webpack plugin code :) Hope it won't break anything.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@ import fs from 'fs';
 import path from 'path';
 import webpack from 'webpack';
 import sources from 'webpack-sources';
+import loaderUtils from 'loader-utils';
 
 const { ConcatSource, SourceMapSource, OriginalSource } = sources;
 const { Template } = webpack;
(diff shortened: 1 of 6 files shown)


developer-lindner:
Documentation should be updated as well!

vertic4l:
Let's get this merged. Need it now :P

vertic4l:
@michael-ciniawsky any time table when this gets merged finally?

michael-ciniawsky:
@vertic4l I don't have access here, that's up to Tobias atm :), From my side it's g2g, the only thing notable here is that the hash of the JS chunk will nevertheless change when the CSS changes...

michael-ciniawsky:
So it mitigates caching problems at best (the CSS will keep the hash)

vertic4l:
@michael-ciniawsky well, i can't actually close my own PR within a project (which updates to webpack 4 setup) as long as this plugin or ExtractTextPlugin doens't work properly with incremenatl build, contentHash and webpack 4. Pretty annoying actually, but can't really help either :/

src/index.js
```diff
@@ -95,6 +96,19 @@ class MiniCssExtractPlugin {
   }
 
   apply(compiler) {
+    // add contenthash support
+    compiler.hooks.emit.tap(pluginName, (compilation) => {
+      const regexp = /\[(?:(\w+):)?contenthash(?::([a-z]+\d*))?(?::(\d+))?\]/ig;
+      Object.keys(compilation.assets).forEach((filename) => {
```
7rulnik: What about `object.entries`?

ooflorent: Not available in node 6.

sokra:
I'll add `[contenthash]` support to the webpack core, as this is also needed for js files when CSS is part of the chunk.

Thanks for your PR anyway.

KyleAMathews:
So would we then want to use `[contenthash]` for js chunks too?

Issue #52, opened 2018-03-23
Question: Multiple JS chunks, single CSS chunk


AvraamMavridis:
Is it possible to have multiple JS chunks but a single css output file?

My current setup is:

```
  optimization: {
    minimize: true,
    minimizer: [
      new UglifyJSPlugin({
        uglifyOptions: {
          compress: {
            passes: 2,
          },
          output: {
            comments: false,
          },
        },
      }),
    ],
    splitChunks: {
      chunks: 'all',
      cacheGroups: {
        vendor: {
          filename: 'vendorbundle.js',
          test: m => /node_modules/.test(m.context),
        },
      },
    },
  },

...
...

    new MiniCssExtractPlugin({
      filename: 'styles.css',
    }),
```

This generates 2 JS files, and 2 css files (styles.css, and 1.styles.css)

What I would like to have is a single css file, is that doable?

Akkenar:
I've been using something like that for the same use-case with success:

```
optimization: {
    splitChunks: {
      cacheGroups: {
        styles: {
          name: 'styles',
          test: /\.s?css$/,
          chunks: 'all',
          minChunks: 1,
          reuseExistingChunk: true,
        }
    }
}
```

AvraamMavridis:
Doesnt seem to work for me

![screen shot 2018-03-23 at 13 11 12](https://user-images.githubusercontent.com/283690/37828492-b9c70476-2e9b-11e8-8821-5446e110744a.png)

AvraamMavridis:
@Akkenar can you share all of the config file please?

Akkenar:
@AvraamMavridis it seems that I was missing an `enforce: true` on the rule.

You'll find an example project there: https://github.com/Akkenar/CSS-Single-File-Example

Webpack config:

```
import webpack from 'webpack';
import path from 'path';
import MiniCssExtractPlugin from 'mini-css-extract-plugin';
import HtmlWebpackPlugin from 'html-webpack-plugin';
import OptimizeCSSAssetsPlugin from 'optimize-css-assets-webpack-plugin';

const vendors = ['react', 'react-dom'];
const isProduction = process.env.NODE_ENV === 'production';

export default {
  name: 'client',
  mode: isProduction ? 'production' : 'development',
  target: 'web',
  entry: {
    bundle: './entry.js',
    vendors,
  },
  context: path.join(__dirname, 'src'),

  stats: {
    // Disable the verbose output on build
    children: false,
  },

[...]

PR #60, opened 2018-03-27
Add documentation in README.md about how to extract all CSS in one file

Akkenar:
Following [discussions on the issues](https://github.com/webpack-contrib/mini-css-extract-plugin/issues/52), this PR aims at documenting the common use-case of extracting all CSS in one file, similar to what https://github.com/webpack-contrib/extract-text-webpack-plugin does.

Tested in a small project there: https://github.com/Akkenar/CSS-Single-File-Example

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,47 @@ The `href` of the `<link>` tag has to match the URL that will be used for loadin
 The `data-href` attribute can be used for `<link>` and `<style>` too.
 When inlining CSS `data-href` must be used.
 
+#### Extracting all CSS in a single file
+
+Similar to what [extract-text-webpack-plugin](https://github.com/webpack-contrib/extract-text-webpack-plugin) does, the CSS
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,6 +76,47 @@ The `href` of the `<link>` tag has to match the URL that will be used for loadin
...
+          test: /\.css$/,
+          chunks: 'all',
+          enforce: true
+        }
+      }
+    }
+  },
+  plugins: [
+    new MiniCssExtractPlugin({
+      filename: "[name].css",
```
evilebottnawi: Can we add `[contenthash]` template to avoid problem with cache who use this configuration

Akkenar: I agree with you, `[hash]` could also help with the cache, or the `hash: true` option on `HtmlWebpackPlugin` if you're using it. I guess it depends on your project setup. 

Not quite sure this documentation section is about cache burst though, hence not adding it in the example configuration.

PR #72, opened 2018-03-29
support incremental build

sokra:
This does NOT include HMR support.

It makes incremental builds possible but doesn't integrate with HMR. So for HMR, CSS will not reload.


src/index.js
```diff
@@ -61,6 +65,17 @@ class CssModule extends webpack.Module {
     return resource;
   }
 
+  updateCacheModule(module) {
+    console.log('updateCacheModule', module.content);
```
evilebottnawi: It is require?

sokra: removed

PR #82, opened 2018-04-03
chore(defaults): Applies latest webpack-defaults

d3viant0ne:
#### General Boilerplate (04d9a02)

 - Migrates to CircleCI 2.0 from Travis
 - Removes Appveyor by default ( can be opted into )
 - Uses a non-deprecated form of eslit config
 - Adds base codecov config
 - Adds commitlint config / hook
 - Updates PR templates & adds Codeowners
 - Prettier tooling & complementary eslint config added

####  package.json meta (12191e9)
 - Updates engines range to disclude nodejs 7 ( we test on 9.x )
 - Moves to scoped version of eslint config w/ prettier support
 - Moves to scoped version of schema-utils with pretty output
 - Adds commit message validation scripts
 - Update CI related scripts for CircleCI 2.0

#### Continuous Integration (94365da)

 - Implements workflow on custom build containers
 - Fails fast on static analysis or minimum node version test failure
 - Executes canary suite ( or doesn't ) based on semver comparison of Webpack dist-tags
 - Fail fast approach running static analysis & unit tests against the minimum supported node version w/ webpack@latest, fanning out to cover the remaining matrix if those two steps pass.

##### Branch / Fork build workflow

screen shot 2018-04-03 at 5 32 57 am

##### Master build workflow

fullbuild-workflow

#### Prettier (629ef38)

 - Applies Prettier style updates ( right click & format with prettier plugin for VSCode / Atom ) will read from the `.prettierrc`
 
_Note:_  I will update the eslint configurations enforcement of `class-methods-use-this`. I'm pretty sure we override that in every one of the contrib libs.


package.json
```diff
@@ -2,71 +2,88 @@
...
     "prebuild": "npm run clean",
     "prepublish": "npm run build",
     "release": "standard-version",
+    "release:ci": "conventional-github-releaser -p angular",
+    "release:validate": "commitlint --from=$(git describe --tags --abbrev=0) --to=$(git rev-parse HEAD)",
     "security": "nsp check",
     "test": "jest",
     "test:watch": "jest --watch",
     "test:coverage": "jest --collectCoverageFrom='src/**/*.js' --coverage",
-    "test:manual": "webpack-dev-server test/manual/src/index.js --open --config test/manual/webpack.config.js",
```
sokra: This script should be kept

PR #101, opened 2018-04-17
Add an advanced configuration example in README

sylv3r:
As discussed with @evilebottnawi in #9, here is a PR adding an advanced example in the docs to avoid confusion and problems with the use of this plugin and the `style-loader`, especially when one wants to have HMR in `development` and its styles extracted for `production` builds.


sylv3r:
Not sure what to do @evilebottnawi ... I already signed the CLA and the email used for the commit is in my github account (not the primary though, maybe that's the problem ?).

PR #134, opened 2018-05-08
dynamic import css chunk support old browser

gaterking:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Additional Info

In old browser, such as android 4.3, link tag not support 4.3. Although these devices is few,  the onload callback resolve function will not be called. This PR ignore onload for these devices and resolve immediately.


evilebottnawi:
@gaterking can you provide minimum reproducible test repo and browser version to ensure it is does not works for all, thanks!

gaterking:
@evilebottnawi http://pie.gd/test/script-link-events/
From this test page,  we can see many old browsers not support link tag onload event.

evilebottnawi:
@gaterking also please accept CLA

evilebottnawi:
@gaterking can you faced this problem ?

evilebottnawi:
@gaterking looks good solution, but i think we should search better solution for detecting this

src/index.js
```diff
@@ -338,8 +338,10 @@ class MiniCssExtractPlugin {
                   ]),
                   '}',
                   'var linkTag = document.createElement("link");',
+                  'var isOldWebKit = Number(navigator.userAgent.replace(/.*AppleWebKit\\/(\\d+)..*/, "$1")) < 536;',
```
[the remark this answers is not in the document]
gaterking: At present, old browser is less. I only use regex to match AppleWebKit 536, bescause I think it will cover most of old browser. There is another solution to check version: 
https://github.com/filamentgroup/loadCSS/blob/master/src/onloadCSS.js
As bellow is another solution to check if css is loaded:
http://www.backalleycoder.com/2011/03/20/link-tag-css-stylesheet-load-event/
https://www.zachleat.com/web/load-css-dynamically/
Another question: Why we need to wait resolve for css chunks is loaded? CSS chunkses are appended in sequence,  they can't impact each other. The waiting will increase js render time inversely. There is so less case we need to caculate element style when the the self css is loaded but next css is not loaded.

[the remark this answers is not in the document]
gaterking: https://github.com/guybedford/require-css/blob/master/css.js
A CSS Loader plugin for RequireJS which can be refered to.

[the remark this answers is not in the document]
gaterking: Yes, it's solid. We can remove some inspections such as ie or low version firefox. Webpack 4 don't support ie6 to 8.

Kovensky:
This also happened to be a blocker for a smartphone-specific project at work. I'll link this to the person involved so he can check if this works 🤔

Kovensky:
Note that it's important to know when a dynamically inserted style is loaded. Not knowing this leads to FOUC, which can be disastrous if you're doing things like `getBoundingClientRect` on component mount if it happens before the load is done.

src/index.js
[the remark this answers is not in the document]
evilebottnawi: @gaterking we can avoid part of code for old `ie`

gaterking:
[index.js](https://raw.githubusercontent.com/gaterking/mini-css-extract-plugin/master/src/index.js)
@evilebottnawi Can you review my new code? If OK, I will create a new PR.
Refer to [RequieJS plugin](https://raw.githubusercontent.com/guybedford/require-css/master/css.js), it use regex to check browser, but it don't check webkit version.
In new solution, I refer to [Dojo plugin](https://www.sitepen.com/blog/2014/07/02/dojo-faq-dynamically-loading-css/) and [xstyle](https://github.com/kriszyp/xstyle/blob/master/core/load-css.js). I use setInterval to check css file is loaded in this case.
I am sure we need to support webkit < 536 only, Becase of  Webpack 4.

src/index.js
```diff
@@ -349,7 +362,25 @@ class MiniCssExtractPlugin {
                     'reject(err);',
                   ]),
                   '};',
-                  'linkTag.href = fullhref;',
+                  '} else {',
+                  'var errorTimeout = 60000;',
```
evilebottnawi: What about is loading more than `60000` ?

gaterking: It's a timeout threshold for css loading. To make sure chunk is resolved, I think we need a stop time to clear setInterval when the css loaded fail.

hxlniada:
may be this is a better solution: https://stackoverflow.com/a/5371426

gaterking:
I think this PR is not necessary for most developer, maybe we can close it. If somebody need to support old browser, I suggest to clone own repo.

lili21:
@evilebottnawi what do you think?

lili21:
@hxlniada It's a neat solution, just need a little change.

```javascript
...
var img = document.createElement('img');
img.onerror = function (e) {
  var request = event && event.target && event.target.src || fullhref;
  var i = document.styleSheets.length;
  while(i--) {
    if (document.styleSheets[i].href === request) {
      return resolve();
    }
  }
  var err = new Error('Loading CSS chunk ' + chunkId + ' failed.\n(' + request + ')');
  err.request = request;
  reject(err);
};
img.src = fullhref;
...
```

gaterking:
#299

src/index.js
```diff
@@ -340,7 +340,20 @@ class MiniCssExtractPlugin {
                   'var linkTag = document.createElement("link");',
                   'linkTag.rel = "stylesheet";',
                   'linkTag.type = "text/css";',
-                  'linkTag.onload = resolve;',
+                  'linkTag.href = fullhref;',
+                  "// old webkit's would claim to have onload, but didn't really support it",
+                  '// https://github.com/kriszyp/xstyle/blob/master/core/load-css.js',
+                  'var webkitVersion = navigator.userAgent.match(/AppleWebKit\\/(\\d+\\.?\\d*)/);',
```
sokra: This extra check affects only 0.02% of the internet. I don't think it's worth including this workaround of everybody.

src/index.js
```diff
@@ -340,7 +340,20 @@ class MiniCssExtractPlugin {
...
+                  'webkitVersion = webkitVersion && +webkitVersion[1];',
+                  'if(linkTag.onload === null && !(webkitVersion < 536)){',
+                  '// most browsers support this onload function now',
+                  'linkTag.onload = function(){',
+                  '// cleanup',
+                  'linkTag.onload = null;',
```
sokra: indent

src/index.js
```diff
@@ -349,7 +362,25 @@ class MiniCssExtractPlugin {
                     'reject(err);',
                   ]),
                   '};',
-                  'linkTag.href = fullhref;',
+                  '} else {',
+                  'var errorTimeout = 60000;',
+                  'var startTime = Date.now();',
+                  'var interval = setInterval(function(){',
+                  'if(linkTag.style){',
+                  'clearInterval(interval);',
```
sokra: indent

src/index.js
```diff
@@ -349,7 +362,25 @@ class MiniCssExtractPlugin {
...
+                  'resolve();',
+                  '}',
+                  'if(!linkTag.style && (Date.now() - startTime) > errorTimeout){',
```
sokra: The support for onload and onerror differ. Make sure to handle them separately. Use a setTimeout for error handling and don't merge it into the setInterval for onload.

src/index.js
```diff
@@ -349,7 +362,25 @@ class MiniCssExtractPlugin {
                     'reject(err);',
                   ]),
                   '};',
-                  'linkTag.href = fullhref;',
+                  '} else {',
+                  'var errorTimeout = 60000;',
+                  'var startTime = Date.now();',
+                  'var interval = setInterval(function(){',
+                  'if(linkTag.style){',
```
sokra: Instead of polling this seem to be a better solution for this problem: https://github.com/webpack-contrib/mini-css-extract-plugin/pull/299/files

Issue #92, opened 2018-04-06
[path] placeholder


glen-84:
`[name]` returns the name of the chunk that imports the CSS file, from what I understand. I need to get the name/path of the imported CSS file.

The file-loader has a `[path]` placeholder that returns `The path of the resource relative to the context`. Is it possible to add something similar to this plugin?

(either that, or add this information to the [filename function](https://github.com/webpack-contrib/mini-css-extract-plugin/issues/67))

sokra:
This doesn't make sense. What's you use case for this?

glen-84:
If I do this:

`scripts/main.ts`
```ts
import "../styles/main.scss";
```

... the CSS file is saved to `scripts/main-57a9c47e715c86dc8071.css`. However, I obviously want CSS to be saved to a location based on the path it was imported from (`/styles` in this case).

This is similar to #67, and if that function has enough data in the arguments provided, it might be a viable alternative.

Either that, or CSS entrypoints 😛 ([related](https://github.com/webpack/webpack/pull/6448))

PS. I'm new to webpack, so if I'm missing something obvious, please excuse me.

glen-84:
@sokra,

I should have mentioned, I use paths for my chunk names, and not simple names like "page1".

By using:

```ts
entry: {
    "scripts/main": "./scripts/main.ts",
    // more
}
```

My JavaScript files are written to subdirectories of the output path (when using `[name]` in the filename template).

If a JS file references a CSS file, and the mini-css-extract-plugin filename template is set to `[name]-[contenthash].css`, the CSS file is written to `scripts/main-{hash}.css`. In other words, `[name]` refers to the chunk name of the entry file.

However, I (and I'm sure others) need to be able to write the CSS to a path based on its actual path on disk (in the src). Either in the form of additional placeholders (this issue), or via a filename function (#67) that is called with enough context to extract that data.

The data seems to be available under `CssModule.issuer.id` (or `.resource`), but I don't really know much about the internals of webpack.

My use case is a multi-page website, where the server-rendered templates (PHP/Twig) should be able to reference an asset using its source path, which then returns the relevant path(s) in the output directory.

For example, if the template asks for `/scripts/main.js`, `/scripts/main-{hash}.js` will be returned (or multiple assets in some cases). If the template asks for `/styles/example.css`, this won't be found because the file is actually under `/scripts`.

[...]

PR #145, opened 2018-05-14
feat: allow filename to be a `{Function}`

### LINKED ISSUE #143: feat: allow the option filename to be a function
Can we allow the filename and chunkFilename passed in trought the constructor to be functions instead of string only?

In our case it would make it easy to toggle the [hashcontent] for some bundles. 
We would pass the option as followed: 
`filename: (chunk)  =>  chunk.name === "NotHashedBundle" ? '[name].css' : '[name].[contenthash].css'`

Might be worth noting that the latest version of webpack also supports something simular to this in their output.
eg: `filename: (data) => data.chunk.name === "NotHashedBundle" ? '[name].js' : '[name].[chunkhash].js'`

berthertogen:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

closes #143 

### Additional Info

Wasn't able to get the tests to run! But wrote a test for the filename-function.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -111,6 +111,11 @@ class MiniCssExtractPlugin {
       options
     );
     if (!this.options.chunkFilename) {
+      if (this.isFunction(this.options.filename)) {
+        throw new Error(
+          `You are required to provide a value for chunkFilename when using a Function for this.options.filename, please specify the [name], [id] or [chunkhash] in this function`
+        );
+      }
       const { filename } = this.options;
       const hasName = filename.includes('[name]');
       const hasId = filename.includes('[id]');
@@ -170,7 +175,7 @@ class MiniCssExtractPlugin {
                   renderedModules,
                   compilation.runtimeTemplate.requestShortener
                 ),
-              filenameTemplate: this.options.filename,
+              filenameTemplate: this.getFilename(chunk, this.options.filename),
               pathOptions: {
                 chunk,
                 contentHashType: NS,
@@ -194,7 +199,10 @@ class MiniCssExtractPlugin {
                   renderedModules,
                   compilation.runtimeTemplate.requestShortener
                 ),
-              filenameTemplate: this.options.chunkFilename,
+              filenameTemplate: this.getFilename(
+                chunk,
+                this.options.chunkFilename
+              ),
               pathOptions: {
                 chunk,
                 contentHashType: NS,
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -111,6 +111,11 @@ class MiniCssExtractPlugin {
       options
     );
     if (!this.options.chunkFilename) {
+      if (this.isFunction(this.options.filename)) {
+        throw new Error(
+          `You are required to provide a value for chunkFilename when using a Function for this.options.filename, please specify the [name], [id] or [chunkhash] in this function`
```
evilebottnawi: Remove this `throw`, it should be done after we integrate `schema` for options

src/index.js
```diff
@@ -366,6 +374,17 @@ class MiniCssExtractPlugin {
     });
   }
 
+  getFilename(chunk, filename) {
+    return this.isFunction(filename) ? filename(chunk) : filename;
+  }
+
+  isFunction(functionToCheck) {
```
evilebottnawi: Remove this and use `typeof something === 'function'`

aaarichter:
looking forward to this :octocat:

berthertogen:
Sorry it took so long, I remove the console logs and made sure the filename function is called only once.

glen-84:
Is there enough information in `chunk` to figure out the path used to reference the CSS file/the path on disk? (see #92)

src/index.js
```diff
@@ -170,7 +170,11 @@ class MiniCssExtractPlugin {
                   renderedModules,
                   compilation.runtimeTemplate.requestShortener
                 ),
-              filenameTemplate: this.options.filename,
+              filenameTemplate: this.getFilename(
+                chunk,
+                this.options.filename,
+                this.options.processedFilename
```
ooflorent: What is this option? There is no documentation about it and it is not used by the test cases.

src/index.js
```diff
@@ -366,6 +380,19 @@ class MiniCssExtractPlugin {
     });
   }
 
+  getFilename(chunk, filename, processedFilename) {
+    if (!processedFilename) {
+      processedFilename = this.isFunction(filename)
```
ooflorent: Use `return` here. Reassigning parameters result in _deopt_.

ooflorent: Could you use `typeof` instead please?

michael-ciniawsky:
Closing in favor of #225 

 Thx :)

PR #149, opened 2018-05-16
fix(index): check if `document` is defined

ericdevries:
This PR contains a:

- [x ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Rendering bundles targetting nodejs rather than a browser still inserts javascript that assumes it will be executed in an environment that has `document` defined.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -319,6 +319,11 @@ class MiniCssExtractPlugin {
               Template.indent([
                 'promises.push(installedCssChunks[chunkId] = new Promise(function(resolve, reject) {',
                 Template.indent([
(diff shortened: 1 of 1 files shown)


evilebottnawi:
Also please accept CLA

evilebottnawi:
@prate3k sorry for delay, problem still exists?

evilebottnawi:
Do not run plugin in non browsers env, if you have problem with SSR please open new issue with reproducible test repo, thanks

PR #155, opened 2018-05-21
Move webpack from devDeps to deps

Jinjiang:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Actually, `webpack` has been imported in https://github.com/webpack-contrib/mini-css-extract-plugin/blob/41347f72aaf5927a4faacce5e0f43e90126b0c17/src/index.js#L4

It may cause some deps error in some particular `npm install` cases imo.

Thanks.

### Additional Info

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
   "dependencies": {
     "@webpack-contrib/schema-utils": "^1.0.0-beta.0",
     "loader-utils": "^1.1.0",
+    "webpack": "^4.4.0",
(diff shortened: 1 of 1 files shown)


evilebottnawi:
@Jinjiang no, webpack is dev and peer deps

Issue #154, opened 2018-05-20
Advanced example is not working


EddyVinck:
I tried to follow the advanced example, but the ``` devMode ? 'style-loader' : MiniCssExtractPlugin.loader, ``` part always seems to use ```style-loader```. 

My scripts
```
"scripts": {
    "test": "test",
    "build": "webpack-dev-server --config webpack.config.js --mode development", // uses style-loader
    "prod": "webpack --config webpack.config.js --mode production" // uses style-loader
  },
```

Am I missing something?

This is my config:
```
const webpack = require('webpack');
const path = require('path');
const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
const MiniCssExtractPlugin = require("mini-css-extract-plugin");
const HtmlWebpackPlugin = require('html-webpack-plugin');
const devMode = process.env.NODE_ENV !== 'production';

module.exports = {
	entry: './src/index',

	output: {
		filename: '[name].bundle.js',
		path: path.resolve(__dirname, 'dist')
	},

	plugins: [
		new MiniCssExtractPlugin({
			// fallback: "style-loader",
			// filename: 'styles/[name].css',
			filename: devMode ? '[name].css' : 'styles/[name].[hash].css',
      chunkFilename: devMode ? '[id].css' : '[id].[hash].css',
		}),
		new UglifyJSPlugin(),
		new HtmlWebpackPlugin({
			title: "Output management",
			template: 'src/assets/html/index.html'
		}),
	],

	module: {
		rules: [
			{
				test: /\.js$/,
				exclude: /node_modules/,
				include: path.resolve(__dirname, "src"),
				loader: 'babel-loader',

				options: {
					presets: ['env']
				}
			},
			{
				test: /\.(png|svg|jpg|gif)$/,
				use: [
					'file-loader'
				]
			},
			{
				test: /\.(woff|woff2|eot|ttf|otf)$/,
				use: [
					{
						loader: 'file-loader',
						options: {
							name: '[name].[ext]',
							outputPath: 'assets/fonts/'
						}
					}
				]
			},
			{
				test: /\.(scss|css)$/,
				use: [
					devMode ? 'style-loader' : MiniCssExtractPlugin.loader,
					{
						loader: 'css-loader'
					},
					{
						loader: 'sass-loader'
					}
				]
			}
		]
	},

	devServer: {
		contentBase: "./dev",
		compress: true,
		port: "9000"
	},
	watchOptions: {
[...]

PR #159, opened 2018-05-23
docs: update advanced example

tiendq:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** #154 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Update advanced example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,11 @@ This plugin should be used only on `production` builds without `style-loader` in
 
 Here is an example to have both HMR in `development` and your styles extracted in a file for `production` builds.
 
-(Loaders options left out for clarity, adapt accordingly to your needs.)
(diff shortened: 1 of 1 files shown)


Tiendq:
@evilebottnawi `mode` makes sense in the advanced example of this plugin (check README file), the example doesn't work if you use `mode` argument with `webpack`, what @eddyvinck used in #154. It works if `NODE_ENV` is used.

michael-ciniawsky:
This is unrelated to `mini-css-extract-plugin` in particular and therefore unnecessarily 'bloats' the plugin README. The fact that `webpack`'s `mode` doesn't rely on `process.env.NODE_ENV` needs to be extrapolated/documented elsewhere e.g in `webpack/webpack.js.org`

PR #232, opened 2018-08-06
ci(jest): run tests in an explicit environment

michael-ciniawsky:
### Type 

- [x] CI

### SemVer

- Fix (🏷 Patch)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,5 @@
 module.exports = {
+  testEnvironment: 'node',
   transformIgnorePatterns: [
     '/node_modules/',
     '<rootDir>/dist/',


evilebottnawi:
@michael-ciniawsky need investigate `x composes-async should compile to the expected result (1023ms)`

michael-ciniawsky:
We ideally should do this is our boilerplate since this will fail for each and every repo, but i would like to eventually overhaul the boilerplate beforehand (simplify it). Yeah the test failure seems to be real, I try to look at it but not familiar with the test here atm :)

evilebottnawi:
@michael-ciniawsky ping me when you investigate

evilebottnawi:
Also better use https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/jest.config.js file to with `"testEnvironment": "node",`. It is allow for future tests switch env (if we need this), more universal way.

michael-ciniawsky:
I didn't find the correct name in the Jest Docs for adding this to the config file, will update

michael-ciniawsky:
Something is really at odds with the CI, all test pass locally (`node >= v10.0.0`) and when I run `npm lint` myself the following doesn't show up for me at all 😕 (only one warning)

```bash
 /home/circleci/project/src/index.js
    8:8   error  Replace `·Template,·util:·{·createHash·}·` with `⏎··Template,⏎··util:·{·createHash·},⏎`                                    prettier/prettier
  100:10  error  Replace `{·dependencies:·[dependency]·},·callback` with `⏎····{⏎······dependencies:·[dependency],⏎····},⏎····callback⏎··`  prettier/prettier
```

evilebottnawi:
@michael-ciniawsky use `prettier`

michael-ciniawsky:
The problem is I don't get anything to fix on my machine 🙃

PR #236, opened 2018-08-08
fix css files load order mismatch #188

ShanaMaid:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

Issue #202, opened 2018-07-05
CSS ordering from shared chunk not deterministic with multi entry points


ferdaber:
Minimal reproduction repo (kind of): https://github.com/ferdaber/css-plugin-ordering-bug
Node: v9.11.1
OS: MacOS v10.12.6

With the latest version v0.4.1, in multiple-entry-points situations, I found that the CSS ordering is not deterministic. I set up a repo that "reproduces" the bug to the best of my ability (before getting stuck in Webpack internals). 

The gist of it is that in cases where a chunk is shared between multiple chunk groups, it's not always the case that the first chunk group actually has the CSS modules that we want to sort on, which results in this line returning `undefined` for the module index:
https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/index.js#L393

Chunk groups:
* `entry-a~vendor`: has `a.js` and `shared_modules/other.js`
* `entry-b~vendor`: has `b.js`, `shared_modules/red.css` and `shared_modules/blue.css`

In my repo, I set it up so that anything in `shared_modules` is split out into a separate chunk, but have the entry points ordered such that the first chunk group contains no CSS modules (`entry-a`'s dependency tree only contains JS files but touches the `shared_modules` chunk group). This leads to the sorting mechanism as mentioned above to return `undefined` for a CSS module's index. 

The output in this repo still ended up having the correct ordering but I think it was because the `chunk.modulesIterable` set had the correct ordering at the time:

Output css/vendor.css:
```css
html {
  background: blue;
}
html {
  background: red;
}
```

Expected css/vendor.css:
```css
html {
  background: red;
}
html {
  background: blue;
}
```

To my understanding, `chunk.modulesIterable` is generated (or added) asynchronously so the ordering is not always guaranteed?

At this point I got stuck trying to get more information to submit to this issue, if a maintainer could offer me some guidance to what a better chunking strategy would be to work around this, or a fix for this, that would be super appreciated - thanks!

dtothefp:
@evilebottnawi I can report a similar issue from the `v4.0.1` release. This is pretty much the exact same behavior I was seeing when upgrading to Webpack 4 and using the `extract-text` plugin.  

 I'm using PostCSS Loader with CSS modules. In development without extract text everything works as expected, although in prod when using extract text I see a duplication of the same compiled CSS module classes in the bundle and a regression in order of some classes. 

This seems to be happening in combination with using the 
- postcss-import module
[...]

PR #130, opened 2018-05-05
Fix CSS ordering for config with multiple entry points

bawjensen:
Hey, this is a rough draft of a fix for #115. The code I wrote is pretty verbose as I didn't want to pull in any dependencies such as lodash or flatten. Let me know what I can do to improve it and get it ready for merging

fixes #115

PR #242, opened 2018-08-13
Fix order of css import

jarandmi:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **metadata update**
- [ ] **test update** 
- [ ] **typo fix**

### Motivation / Use-Case
Modules with undefined ID will now be sorted to the bottom of array, and imported after the once with and ID

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -387,10 +387,12 @@ class MiniCssExtractPlugin {
     // with different order this can lead to wrong order
     // but it's not possible to create a correct order in
     // this case. Don't share chunks if you don't like it.
+
+
     const [chunkGroup] = chunk.groupsIterable;
     if (typeof chunkGroup.getModuleIndex2 === 'function') {
       modules.sort(
-        (a, b) => chunkGroup.getModuleIndex2(a) - chunkGroup.getModuleIndex2(b)
+        (a, b) => (chunkGroup.getModuleIndex2(a) || 0) < (chunkGroup.getModuleIndex2(b) || 0) ? 1 : -1
       );
     } else {
       // fallback for older webpack versions
--- a/src/index.js
+++ b/src/index.js
@@ -387,8 +387,6 @@ class MiniCssExtractPlugin {
     // with different order this can lead to wrong order
     // but it's not possible to create a correct order in
     // this case. Don't share chunks if you don't like it.
-
-
     const [chunkGroup] = chunk.groupsIterable;
     if (typeof chunkGroup.getModuleIndex2 === 'function') {
       modules.sort(
--- a/src/index.js
+++ b/src/index.js
@@ -5,7 +5,10 @@ import webpack from 'webpack';
 import sources from 'webpack-sources';
 
 const { ConcatSource, SourceMapSource, OriginalSource } = sources;
-const { Template, util: { createHash } } = webpack;
+const {
+  Template,
+  util: { createHash },
+} = webpack;
 
 const NS = path.dirname(fs.realpathSync(__filename));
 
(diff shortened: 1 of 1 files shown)


sokra:
I assume this happens when chunks are shared between multiple chunk groups. This can happen when using `splitChunks` with a constant `name`. Maybe it can be fixed by using all chunk groups in `chunk.groupsIterable`. But make sure to create a test case to reproduce this bug first.

jarandmi:
@sokra Yes, you're right. Thats when it happens. I'm not so familiar with Webpack and groupsIterable, but will try to dig into it.

src/index.js
```diff
@@ -390,7 +398,11 @@ class MiniCssExtractPlugin {
     const [chunkGroup] = chunk.groupsIterable;
```
sokra: Here only the first chunkGroup is used. Maybe this need to be reconsidered. See also comment above.

jarandmi: Yeah, i'm trying to look into this. IS there a reason why the order is sorted by `getModuleIndex2(module)`and not `module.id`?

jarandmi: My issue get solved if i order after module.id.

jarandmi: I will try to make a testcase for it.

jarandmi:
@sokra : I have created a test case: https://github.com/jarandmi/scss-order-bug
Try to run `yarn start:dev`

I think the problem has too do with sass, and the `@import "../variables";``
When i use this the order of the import is in wrong order. The order is also right when i run the webpack-dev-server in production mode.

jarandmi:
It will also be fixed if i sort bu module.id, instead of `chunkGroup.getModuleIndex2(module)`
`modules.sort( (a, b) =>  a.id <  b.id  ? -1  : 1 );`

rowanoulton:
Hi there, dropping in to say we're experiencing this problem as well. `chunkGroup.getModuleIndex2` will return `undefined` for any module that isn't part of the graph of the chosen chunk group. Depending on the location of these `undefined` indexes, the sort may or may not work properly.

Fixing the comparator function to sort `undefined` to the bottom is a good step. I would also encourage we alert the engineer to the situation (eg. `console.warn`) as it'll save people time they would otherwise spend investigating why their CSS order is wrong. 

That being said, I think there's a bigger problem to consider: when sharing CSS between multiple chunk groups, we don't know for sure what the correct order is. Consider the original problem (#130):

a.js
```
import './a.css';
import './b.css';
```

b.js
```
import './b.css';
import './a.css';
```

If we want to share CSS between these two chunks, we have to decide how to order `a.css` and `b.css`. This is a problem even if we fix the sorting of `undefined`. If we use the first entrypoint as the source of truth (as it's currently implemented) our CSS order will be implicitly dependent on entrypoint order. This is not obvious and, I think, fiendishly difficult to debug.

I'm not sure what the proper solution is. It seems to me that the current configuration doesn't provide enough information to correctly judge what CSS order is correct — at least in the case where it's shared across multiple chunks. The remedy might be more configuration, or handing off the responsibility to the user:

```
plugins: [
  new MiniCSSExtractPlugin({
    sharedSortComparator: (a, b) => {
      /* .. your own logic to determine CSS order ... */
    }
  })
]
```

I'd be interested to hear your thoughts on this @jarandmi @sokra. Happy to work on an implementation too. For now we've forked the plugin to revert to the old sort method (ie. using `module.index2` instead of `chunkGroup.getModuleIndex2`).

Possibly related issues: #202, #236

sokra:
Could you check if #246 fixes your problem?

PR #246, opened 2018-08-20
fix(order): use correct order when multiple chunk groups are merged

sokra:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

see #242

Order of modules is incorrect/undefined when merging multiple chunk groups into a single chunk.

This PR also emits a warning when order is conflicting between chunk groups.

### Breaking Changes

bugfix

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -167,6 +167,7 @@ class MiniCssExtractPlugin {
             result.push({
               render: () =>
                 this.renderContentAsset(
+                  compilation,
                   chunk,
                   renderedModules,
                   compilation.runtimeTemplate.requestShortener
@@ -192,6 +193,7 @@ class MiniCssExtractPlugin {
             result.push({
               render: () =>
                 this.renderContentAsset(
+                  compilation,
                   chunk,
                   renderedModules,
                   compilation.runtimeTemplate.requestShortener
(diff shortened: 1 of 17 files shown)


jarandmi:
Yes! This fix my problem. But i'm not sure if we should warn in the console. Do we want the user to take any action to fix this, if shared chunks is used in different order?

sokra:
> Do we want the user to take any action to fix this, if shared chunks is used in different order?

Yes, the user should make sure to reorder the imports to avoid conflicting order.

jarandmi:
I agree, but it will be very hard to use reusable components (like in react) this way.

Example:

**Text.js**
import "./textStyle.css"

**ComponentOne**
import "./componentOneStyle.css"
import Text from "./Text.js"

**ComponentTwo**
import "./ComponentTwo.css"
import Title from "./Title.js"
import Text from "./Text.js"

I think it will be almost impossible to order this the right way if a component is reused in many other components. What do you think?

robwierzbowski:
Just starting to dig through this code. Does the comparison function check index of the import, like this: 

**file 1**

```
import "./a.css"; // 1
import "./b.css"; // 2
import "./d.css"; // 3
```

**file 2**

```
import "./a.css"; // 1
import "./b.css"; // 2
import "./c.css"; // 3
import "./d.css"; // 4, fire warning because d is not index 3, although it's still after a and b
```

or comparative order, like this? 

**file 1**

```
import "./a.css"; 
import "./b.css"; 
import "./d.css";
```

**file 2**

```
import "./a.css"; // before b and d
import "./b.css"; // after a, before d
import "./c.css"; // new
import "./d.css"; // after a and b
```

robwierzbowski:
After reading some of the comments in https://github.com/webpack-contrib/mini-css-extract-plugin/issues/250 I think my team will silence warnings. Apologies for the noise.

PR #265, opened 2018-09-07
refactor: rename `NS` to `MODULE_TYPE` and use a static value

### LINKED ISSUE #263: Question for the authors
Hey guys, I wanted to ask about the reasoning behind using a path for NS variable, instead of something static, like a string. 

Heres the issue, lets say i have a build toolchain. The toolchain contains an import to mini-css, then elsewhere in the project  core i also import mini-css. 

What happens is i get an error saying this[NS] cant be resolved. When logging i discover that each instance has a new name which is used for loader context. 

So the loader is looking for `this[‘plugin-sass/node_modules/extract’]`, but the plugin is providing `this[‘react-static/node_modules/extract']`

I can change this by hard coding a string, and everything works perfectly across the instances. 

`const NS = '_extractCssChunks'; // works perfectly`

Id like to know if there is any particular reason it was coded in this manner?
if i were to change the `NS` to something static in my project, would there any ramifications?

https://github.com/webpack-contrib/mini-css-extract-plugin/blob/6323e178755170c08598f9fe1c8ef8e3cd13154b/src/index.js#L13

Appreciate your work and time.

@sokra

michael-ciniawsky:
#263 

### Typ

- [x] Refactor

### Issues

- Fixes #263 

### SemVer

- [x] Patch

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,3 @@
-import fs from 'fs';
-import path from 'path';
-
 import webpack from 'webpack';
 import sources from 'webpack-sources';
 
@@ -10,7 +7,7 @@ const {
   util: { createHash },
 } = webpack;
 
-const NS = path.dirname(fs.realpathSync(__filename));
+const MODULE_TYPE = 'css/mini-extract';
 
 const pluginName = 'mini-css-extract-plugin';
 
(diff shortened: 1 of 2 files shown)


evilebottnawi:
@michael-ciniawsky maybe patch release?

ulivz:
❤️

Issue #234, opened 2018-08-08
Feature: support minSize


PanJiaChen:
![image](https://user-images.githubusercontent.com/8121621/43824267-15d4162e-9b24-11e8-9b28-d3ac21d9719c.png)

Currently the css inside the bundle will be extracted into a single `.css` file.

But sometimes the css file will be small, only a few kb, but it will cost an additional http request.

Is it better to set a  minSize?

flyyang:
any update?

PR #268, opened 2018-09-08
feat(index): add `minSize` option (`options.minSize`)

manniL:
# Stale!

This PR contains a:

- [x] new **feature**
- [x] **test update** 
### Motivation / Use-Case

**Attention**: This is my first webpack-related PR so please bear with me :see_no_evil:
Also, this seemed way too easy to implement :thinking: 

This PR implements a minimum size (`minSize`) for CSS files to get extracted.

Will add documentation as soon as the PR is mature :relaxed: 

Related: #234


michael-ciniawsky:
There are a few gotchas with this approach

- How to apply the CSS ? Use `<style></style>` similiar to the `style-loader` ?
- This functionality is already covered by [`optimization.splitChunks`](https://webpack.js.org/configuration/optimization/#optimization-splitchunks) (`SplitChunksPlugin`) in a more generic way, besides not extracting small amounts of CSS at all (proposed by the PR)

manniL:
Yes, I'd apply the CSS "inline".

Let me outline a scenario where this feature should have a large impact:

I'm using Vue as Frontend framework and have (little) *component-based styles*. I also have a *global style* file (imagine sth. like bootstrap).

When I now code-split my application, the component JS files (with the component CSS included) will be loaded asynchronously. Usually, the component CSS isn't very large as most styling is done in the large global file.

There are a few scenarios now:

1. Don't extract CSS at all. 
    Bad, as I want to have long term caching for my *global style*.

2. Extract all CSS files. 
    Bad, as I'll have a CSS file per component with a few (say ~5-25) lines of CSS in it, which are even smaller than the HTTP headers of the request to retrieve the file.

3. Extract all CSS and merge into one file (with `splitChunks`)
    Best (current) option, but still including lots of unneeded CSS on start

4. (UPCOMING) Extract only large portions of CSS (in their own files) and leave small CSS in the components

michael-ciniawsky:
kk, applying the inline CSS should ideally work out-of-the-box, similiar to how it currently [works for `<link href="async.chunk.css" rel="stylesheet">`](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/index.js#L323-L371) but via a `<style>` tag, if `content.length < minSize`

michael-ciniawsky:
Since this is likely a bit more involved to implement, it's maybe better to wait for general feedback first before starting to work on it :)

manniL:
Friendly ping 🙌

lsycxyj:
In my opinion, rather than only using `minSize` as the criterion to determine the extracting behavior, I suppose it's better to use something like a test function, since it should be much more flexible to control.

joaumg:
Friendly ping @manniL.

@lsycxyj this can be extended later... also, I don't see any type of function-ish scenarios here...

lsycxyj:
@joaumg Well... I don't think so.  
For instance, I don't think the browser needs to make extra an request to fetch the css file of its own async js chunk which it has been loaded. So I want to disable it and I need a test funciton to test if the processing chunk is an async one.

manniL:
Closing due to inactivity

Issue #278, opened 2018-09-18
Not recovering from failed dynamic import


cwalten:
This issue arises when a dynamic import fails (in this case due to network instability). If the request does fail, the import code will never attempt to request the CSS again. Thus, we will never successfully complete the import. The code which is making that request would be something like the following:
```
import(
  /* webpackChunkName: "AsyncBundle1", webpackMode: "lazy" */
  './AsyncBundle1'
)
.then(AsyncBundle1 => {
  this.AsyncBundle1 = AsyncBundle1.default;
})
.catch((error) => {
  console.log(error);
})
```
Sorry I could not provide a better base example, as this issue requires a server and network manipulation to achieve.

Reproduction steps:

1. The bundle is requested and the network call fails.
2. The CSS request Promise is rejected [here](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/v0.4.2/src/index.js#L353), but the bundle's key in installedCssChunks is not cleared.
3. We retry the request, but [this line](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/v0.4.2/src/index.js#L321) is executed, placing the rejected Promise in the promises array, not a new Promise.
4. This will evaluate to Rejected every time, never even making a second service request for that CSS.

This shows that the installedCssChunks has held onto the Rejected Promise:

<img width="503" alt="asyncbundle1promise" src="https://user-images.githubusercontent.com/19825717/45702034-5dbb8a80-bb36-11e8-8101-33c4c33cf199.png">

This shows that the CSS was never requested again, failing the Promise in the example above (Note: the blank row is a JSON request not relevant for this discussion):

<img width="580" alt="nocssrequest" src="https://user-images.githubusercontent.com/19825717/45702953-ab38f700-bb38-11e8-896b-3a888a954822.png">

This shows the Rejected Promise being pushed into the promises array for 'execution' (immediate rejection):

<img width="662" alt="pusingintopromises" src="https://user-images.githubusercontent.com/19825717/45702036-5dbb8a80-bb36-11e8-9116-12f54b13b0bb.png">

It is possible I have implemented this dynamic request incorrectly, but I was unable to find any solutions which fixed this issue through Webpack or this plugin specifically. Since the Javascript is being requested successfully (see above), it feels like an issue with the template code provided by this plugin.

cwalten:
[...]

PR #292, opened 2018-10-04
fix(index): update to allow requesting failed async css files

cwalten:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
Related to #278 

Since a rejected promise is left in installedCssChunks on a failed async import, the code to add the HTML link tag again is totally skipped.

Further, the code which has been removed in this PR is currently causing dynamically imported CSS files which have failed (but still have the link/ style tag added in the DOM) to simply return a resolve() instead of retrying the request. If we return resolve() simply because the link tag is in the DOM, the request is never retried and, though the Promise resolves correctly, the CSS is not loaded into the page.

### Breaking Changes
Not sure, but I do not believe so.

### Additional Info
These changes are modeled off of the async jsonp template code from webpack found [here](https://github.com/webpack/webpack/blob/9b97f3aeb124776731b9830c4964f496a3cdfed8/lib/web/JsonpMainTemplatePlugin.js#L276), where they do not do these duplication checks and reset the chunk key in the installedChunks object when the request fails.


cwalten:
the failed test is also failing in master [here](https://circleci.com/gh/webpack-contrib/mini-css-extract-plugin/643)

tbiethman:
@evilebottnawi @cwalten Any updates regarding the status of this PR? I would like to start consuming these changes when possible.

test/manual/index.html
```diff
@@ -42,6 +42,11 @@
     <p>Preloaded inlined CSS: Must be green.</p>
     <p><button class="preloaded-button2">Pressing this button</button> displays an alert and should turn red.</p>
   </div>
+  <div class="test lazy-failure-css">
```
ooflorent: Could you move this block below `.lazy-css`?

cwalten: Sure, moved it up under `.lazy-css`

src/index.js
```diff
@@ -359,6 +343,7 @@ class MiniCssExtractPlugin {
                     'var request = event && event.target && event.target.src || fullhref;',
                     'var err = new Error("Loading CSS chunk " + chunkId + " failed.\\n(" + request + ")");',
                     'err.request = request;',
+                    'delete installedCssChunks[chunkId]',
                     'reject(err);',
```
sokra: Maybe also remove the failed style tags from the document.

src/index.js
```diff
@@ -334,22 +334,6 @@ class MiniCssExtractPlugin {
...
-                    'if(tag.rel === "stylesheet" && (dataHref === href || dataHref === fullhref)) return resolve();',
-                  ]),
-                  '}',
-                  'var existingStyleTags = document.getElementsByTagName("style");',
-                  'for(var i = 0; i < existingStyleTags.length; i++) {',
-                  Template.indent([
-                    'var tag = existingStyleTags[i];',
-                    'var dataHref = tag.getAttribute("data-href");',
-                    'if(dataHref === href || dataHref === fullhref) return resolve();',
-                  ]),
```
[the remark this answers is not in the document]
cwalten: thanks for the info, I have added the code back in

src/index.js
cwalten: I have added code to remove the tag should onError be called, thanks for the suggestion

goenning:
Just for the record, I replaced my mini-css-extract-plugin dependency with cwalten's master and it solved the problem. I can now retry loading CSS files just like it's done with JS files. 👍

Issue #294, opened 2018-10-11
link element load event not fired on old browser.


lili21:
`MCEP` use `load` event to resolve the promise, while the event is not supported well on some old browser, like Android 4.3 webview.

It breaks the **webpack** `dynamic import` feature, cause the promise will never be resolved.

```js
// index.js
import('./app.js').then(() => {
  // the function will never be called
  console.log('App Loaded')
})
```

```js
// app.js
import './app.css'
console.log('Hello, app')
```

[Reproduce repo](https://github.com/lili21/mcep-dyamic-import-issue)

lili21:
@evilebottnawi I don't think you understand the issue. maybe it's my fault, my english isn't very good.

It's not about Promise polyfills, and it's not the async `js` file issue. If you remove the `import './app.css'` code, it will works as expecting.

and I think it's the issue relate to this repo. as I said, the plugin using `load` event to detect the link be loaded and resolve the promise, but `load` event isn't supported well.

Please reopen it.

evilebottnawi:
@lili21 can you create minimum reproducible test repo and affected browsers?

lili21:
Yeah. [here you go](https://github.com/lili21/mcep-dyamic-import-issue)

I only tested it on Android mobile. Android 4.4+ is good. Android 4.3 isn't.

evilebottnawi:
@lili21 Looks related https://github.com/webpack-contrib/mini-css-extract-plugin/pull/134, but solution is not very good, wee need search better, feel free to investigate

moyus:
same error here. When use webpack dynamic `import()`, all created css is not loaded in old android 4.3 browsers, which lead to async module not resolved.

lili21:
You can extract all css in a single file to prevent this, But it's bad for performance.

Jessidhia:
I present to you this horribleness as a workaround, which does work on Chrome 14, one of the affected browsers according to https://pie.gd/test/script-link-events/:

![image](https://user-images.githubusercontent.com/73085/51379588-23e5a980-1b53-11e9-9c83-61da0837a877.png)

I did not _actually_ test this with `mini-css-extract-plugin`, but I did write the same code as the one `mini-css-extract-plugin` would try to run when you `import()` an extracted file. The important code is inside `function polyfill()`:

```js
// @ts-check
import 'core-js/fn/promise'
// yep, Chrome 14 didn't have this either
import MutationObserver from 'mutation-observer'

polyfill()
[...]

PR #299, opened 2018-10-20
fix(index): dynamic css loading support for older browsers (#294)

moyus:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

This PR solve the issue:`link` element load event not fired on old browser #294 by using `onerror` event of `img` element if user's browser is old enough.

### Breaking Changes

None

### Additional Info

None


ooflorent:
This build of WebKit is the one (and previous) shipped with iOS 6. Do we really need to support something that old?

moyus:
There are still 3% mobile devices using Android 4.3 and below. Yes, the number seems small, however when those users contact you and complain about the app is not working on his phone, coders still need to fix it. Every user matters. 😀

Issue #73, opened 2018-03-30
Module build failed: TypeError: this[NS] is not a function


felipero:
After a fresh install of webpack 4.4.1 and mini-css-extract-plugin I'm getting this error:

```
  ERROR in   Error: Child compilation failed:
  Module build failed: TypeError: this[NS] is not a function

  - loader.js:147 childCompiler.runAsChild
    [kickante_front]/[mini-css-extract-plugin]/dist/loader.js:147:15
```

My configuration file looks like this:

```js
const MiniCssExtractPlugin = require("mini-css-extract-plugin");
const HtmlWebpackPlugin = require('html-webpack-plugin');

module.exports = (env = {}) => {
  const isProduction = env.production === true;

  return {
    plugins: [
      new MiniCssExtractPlugin({
        filename: '[name].css',
        chunkFilename: "[id].css"
      }),
      new HtmlWebpackPlugin({
        title: 'wp4test',
        template: 'src/index.html',
        minify: {
          html5: true,
          removeComments: isProduction,
          collapseWhitespace: isProduction,
          preserveLineBreaks: true,
          decodeEntities: true,
        },
      })
    ],
    module: {
      rules: [
        {
          test: /\.(scss|sass|css)$/i,
          use: [MiniCssExtractPlugin.loader, "css-loader"],
        },
        {
          test: /\.(png|jpg|gif)$/i,
          use: [
            {
              loader: 'url-loader',
              options: {
                limit: 10000,
                outputPath: '/images/',
                name: '[name].kick.[hash:8].[ext]',
              },
            },
          ],
        },
      ]
    }
  }
};
```
The line with error is this: 
```js
this[NS](text);
```
And the value of NS is:

```
/Users/felipero/workspace/learning_experiments/wp4test/node_modules/mini-css-extract-plugin/dist
```
Any ideas of what it could be?

nicksrandall:
I'm not sure if this will help but I was getting this error only on my dev builds because I was only adding the plugin on prod even though both prod and dev were both using the loader.

felipero:
@nicksrandall what do you mean by adding to prod? Do you mean adding to "dependencies" instead of "devDependencies"?

If so, I tried here and it didn't work.

felipero:
I found out something that It might help. `this[NS]` is undefined when requiring from HTML, like that: `<link rel="stylesheet" type="text/css" href="<%=require('./main.css')%>" />`

[...]

PR #318, opened 2018-11-26
Avoid `this[MODULE_TYPE] is not a function` error

MoOx:
This PR contains a:

- [x] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Following https://github.com/webpack-contrib/mini-css-extract-plugin/issues/73#issuecomment-441717981

### Breaking Changes

No

### Additional Info

I didn't add a test because I made this fix from GitHub UI :)

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -124,7 +124,7 @@ export function pitch(request) {
           };
         });
       }
-      if(typeof this[MODULE_TYPE] !== "function") {
+      if(typeof this[MODULE_TYPE] !== 'function') {
         return callback(new Error("It seems you are using the loader but forgot to register the plugin"));
       }
       this[MODULE_TYPE](text);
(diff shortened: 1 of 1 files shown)


src/loader.js
```diff
@@ -124,6 +124,9 @@ export function pitch(request) {
           };
         });
       }
+      if(typeof this[MODULE_TYPE] !== "function") {
```
naoisegolden:
```suggestion
      if(typeof this[MODULE_TYPE] !== 'function') {
```

src/loader.js
```diff
@@ -124,6 +124,9 @@ export function pitch(request) {
...
+        return callback(new Error("It seems you are using the loader but forgot to register the plugin"));
```
naoisegolden:
```suggestion
        return callback(new Error('It seems you are using the loader but forgot to register the plugin'));
```

alexander-akait:
Sorry for delay, fixed in master since v2

PR #321, opened 2018-12-06
Feat/filename transform

kevinchappell:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

The `filename` option currently takes a string the enables devs to customize the filename of the generated css file. This works well and the customized filename is created in the same output directory as the js file it is extracted from, however there are cases when the generated files may need to be in different directories. This change enables modifying the `filenameTemplate` by providing `chunkhash`, `contenthash`, `name`, and `id` as arguments when a function is used for the `filename` option instead of a string.

#### Use Case:
webpack.config.js - `entry`
```javascript
  entry: {
    'dist/main': 'src/index.js',
    'demo/assets/js/demo': 'src/demo/index.js',
  }
```
The above config will create `dist/main.js` and `demo/assets/js/demo.js`.

#### Use:
webpack.config.js - `plugins`
```javascript
  plugins: [
    new MiniCssExtractPlugin({
      filename: ({ name }) => `${name.replace('/js/', '/css/')}.min.css`,
    })
  ]
```
We can change the output directory, chunkhash length and more by using a function that gets called during `compilation.mainTemplate.hooks.renderManifest`.

### Breaking Changes

Most likely none

### Additional Info
Improved some existing tests. Now testing filename and path generation.


evilebottnawi:
What is use case? Why it is need if you have `output` as `Function` in webpack?

pgoforth:
> What is use case? Why it is need if you have `output` as `Function` in webpack?

Exporting your CSS to a different subfolder based on the chunk.

test/cases/filename/webpack.config.js
```diff
@@ -0,0 +1,24 @@
+const Self = require('../../../');
+
+module.exports = {
+  entry: {
+    'demo/js/main': './index.js',
```
pgoforth: Should fix your test fail
```suggestion
    'expected/demo/js/main': './index.js',
```

evilebottnawi:
@pgoforth 
Why don't use https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L857?

pgoforth:
> @pgoforth
> Why don't use https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L857?

@evilebottnawi 
That handles the JS files nicely, and supports having a function as a callback. The MiniCSSExtractPlugin, however, currently does not support providing a function as a callback, therefore the CSS will be extracted and placed into the same directory as your JS file. Most times this is acceptable, however, if you have a desired folder structure like the following:
```
dist
│   
└───app1
│   │
│   └─  app1.js
│   └─  app1.css
│   
└───app2
│   │
│   └─  app2.js
│   └─  app2.css
│
└───cmn
│   └───css
│   │   │
│   │   └─  vendor.css
│   │
│   └───js
│   │   │
│   │   └─  vendor.js
```
...it is easy to get the `vendor.js` file to go into `cmn/js`, but the current MiniCSSExtractPlugin would generate the `vendor.css` file into the `cmn/js` directory. Not always desireable.

TimothyBJacobs:
We need this functionality as well.

domonique-ncino:
Can I help in anyway to get this pushed faster? I need this functionality badly.

evilebottnawi:
Need rebase and fix CI problem

src/index.js
```diff
@@ -112,21 +114,18 @@ class MiniCssExtractPlugin {
...
       options
     );
     if (!this.options.chunkFilename) {
       const { filename } = this.options;
-      const hasName = filename.includes('[name]');
-      const hasId = filename.includes('[id]');
-      const hasChunkHash = filename.includes('[chunkhash]');
       // Anything changing depending on chunk is fine
-      if (hasChunkHash || hasName || hasId) {
+      if (typeof filename === 'string' && REGEXP_PLACEHOLDERS.test(filename)) {
```
evilebottnawi: Don't use regexp , it is bad for perf

kevinchappell: Normally I agree but in this case it was faster than 3 `String.includes` http://jsben.ch/LMM6o

evilebottnawi: @kevinchappell thanks for link, looks good :+1:

evilebottnawi:
@kevinchappell need fix CI problem

src/index.js
```diff
@@ -112,21 +114,18 @@ class MiniCssExtractPlugin {
...
         this.options.chunkFilename = filename;
       } else {
         // Elsewise prefix '[id].' in front of the basename to make it changing
-        this.options.chunkFilename = filename.replace(
+        this.options.chunkFilename = DEFAULT_FILENAME.replace(
```
sokra: Should use `filename` when it's a string here.

src/index.js
```diff
@@ -170,6 +169,10 @@ class MiniCssExtractPlugin {
           const renderedModules = Array.from(chunk.modulesIterable).filter(
             (module) => module.type === MODULE_TYPE
           );
+          const { filename } = this.options;
+          const filenameTemplate =
```
sokra: You can pass a function as filenameTemplate.

test/TestCases.test.js
```diff
@@ -60,20 +60,37 @@ describe('TestCases', () => {
...
-              path.resolve(expectedDirectory, file),
-              'utf-8'
-            );
-            const actualContent = fs.readFileSync(
-              path.resolve(outputDirectoryForCase, file),
-              'utf-8'
+
+          for (const file of walkSync(expectedDirectory)) {
+            const actualFilePath = file.replace(
+              new RegExp(`/cases/${directory}/expected/`),
```
sokra: This doesnt't work on windows, or if directory contains special chars.

sokra: If it fails to replace it will compare the file with itself making the test always pass.

README.md
```diff
@@ -268,6 +268,17 @@ module.exports = {
 }
 ```
 
+#### Filename as function instead of string
+
+By using a function instead of a string, you can use chunk data to customize the filename. This is particularly useful when dealing with multiple entry points and wanting to get more control out of the filename for a given entry point/chunk. In the example below, the we'll change the filename to output the css to a different directory.
+
+```javascript
+const miniCssExtractPlugin = new MiniCssExtractPlugin({
+  filename: ({ name, chunkhash }) =>
```
sokra: This should use the same arguments as function as `output.filename` in webpack.

test/TestCases.test.js
sokra: Maybe use `path.relative` instead.

src/index.js
kevinchappell: Can you elaborate more on this?

kevinchappell: oh, do you mean a separate option as `filenameTemplate` instead of `filename` for both?

test/TestCases.test.js
kevinchappell: good catch, fixed. I used `path.resolve` instead though because the output file can be in a nested directory.

src/index.js
kevinchappell: `filename` will not be a string here.

test/TestCases.test.js
kevinchappell: @sokra Just pushed an update, can you verify the `actualFilepath` works on windows now?

src/index.js
kevinchappell: actually i guess it could be. will change

domonique-ncino:
I'm not familiar with webpack plugin development. If I can get pushed in the right direction to learn so this can get pushed would be great. This is a major need.

src/index.js
kevinchappell: :man_facepalming: I see what you meant. fixed

test/TestCases.test.js
kevinchappell: now using `path` to correctly resolve the directories

kevinchappell:
Are anymore changes needed on this?

src/index.js
kevinchappell: Switched to filenameTemplate

PR #328, opened 2018-12-14
Some small optimizations for stylesheet injection

developit:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** _(not yet!)_
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

There are a few cases where the code injected into a bundle could be a bit smaller. I've addressed two really simple ones here - removing the `type` property from `<link rel=stylesheet>`, and using `document.head` ([IE9+](https://developer.mozilla.org/en-US/docs/Web/API/Document/head)) instead of `getElementsByTagName('head')[0]`.

### Breaking Changes

This requires IE9+. Not sure if there's a browser support target for mini-css-extract-plugin that makes this prohibitive.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -353,7 +353,6 @@ class MiniCssExtractPlugin {
                   '}',
                   'var linkTag = document.createElement("link");',
                   'linkTag.rel = "stylesheet";',
-                  'linkTag.type = "text/css";',
                   'linkTag.onload = resolve;',
                   'linkTag.onerror = function(event) {',
                   Template.indent([
@@ -368,7 +367,7 @@ class MiniCssExtractPlugin {
                   'linkTag.href = fullhref;',
                   crossOriginLoading
                     ? Template.asString([
-                        `if (linkTag.href.indexOf(window.location.origin + '/') !== 0) {`,
+                        `if (linkTag.href.indexOf(location.origin + '/') !== 0) {`,
                         Template.indent(
                           `linkTag.crossOrigin = ${JSON.stringify(
                             crossOriginLoading
@@ -377,8 +376,7 @@ class MiniCssExtractPlugin {
                         '}',
                       ])
                     : '',
-                  'var head = document.getElementsByTagName("head")[0];',
-                  'head.appendChild(linkTag);',
+                  'document.head.appendChild(linkTag);',
                 ]),
                 '}).then(function() {',
                 Template.indent(['installedCssChunks[chunkId] = 0;']),


src/index.js
```diff
@@ -353,7 +353,6 @@ class MiniCssExtractPlugin {
                   '}',
                   'var linkTag = document.createElement("link");',
                   'linkTag.rel = "stylesheet";',
-                  'linkTag.type = "text/css";',
```
evilebottnawi: Why we need remove this? Not sure it is right idea

src/index.js
```diff
@@ -368,7 +367,7 @@ class MiniCssExtractPlugin {
                   'linkTag.href = fullhref;',
                   crossOriginLoading
                     ? Template.asString([
-                        `if (linkTag.href.indexOf(window.location.origin + '/') !== 0) {`,
+                        `if (linkTag.href.indexOf(location.origin + '/') !== 0) {`,
```
evilebottnawi: Always need `window`, you can have scope problem in some rare cases

src/index.js
```diff
@@ -377,8 +376,7 @@ class MiniCssExtractPlugin {
                         '}',
                       ])
                     : '',
-                  'var head = document.getElementsByTagName("head")[0];',
-                  'head.appendChild(linkTag);',
+                  'document.head.appendChild(linkTag);',
```
evilebottnawi: Hm IE9 doesn't support `getElementsByTagName`?

src/index.js
evilebottnawi: Removing `text/css` good for HTML5, but some browsers (include mobile) still have bad supporting HTML5

src/index.js
ScriptedAlchemy: @evilebottnawi looks safe to me?

https://caniuse.com/#search=getElementsByTagName

evilebottnawi: @ScriptedAlchemy for me too, so i asked question why it was changed

ScriptedAlchemy: Yeah good point. Though both of them do work with old I.E. 

However I’m unsure how widely supported ‘document.head’ is especially in older browsers

src/index.js
developit: The default value of the `type` attribute has always been `text/css`. The only difference in HTML5 was making it pass validation when omitted. There is no functionality change with regard to this behaviour between the two specs.

You can test this yourself - load this webpage in IE6:

```html
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html lang="en">
  <head>
    <title>HTML 4 Strict</title>
    <link rel="stylesheet" href="style.css">
    <!-- make sure style.css exists and and is not empty -->
  </head>  
  <body>
    <script type="text/javascript">
      alert(document.styleSheets.length);
    </script>
  </body>
</html>
```

alert 1 in IE6

src/index.js
developit: `document.head` is supported in IE9+.
https://developer.mozilla.org/en-US/docs/Web/API/Document/head#Browser_compatibility

src/index.js
developit: What scope problem are you referring to?

`var location = 1` or `location=1` in global scope redirect the page, as does `window.location=1` - it's exceptionally unlikely the value would be redefined. It could be shadowed, but that's equally true for any variable including `window`.

developit:
Just to clarify my intention with this PR:

The code generated here is injected into a large percentage of websites. Each byte saved represents a nontrivial reduction in bundle size across the whole web.

src/index.js
jakearchibald: Note that the code already uses `document` rather than `window.document`.

ScriptedAlchemy: Fair point... we should side with implicit or explicit conventions throughout

src/index.js
ScriptedAlchemy: Nicely done

src/index.js
ScriptedAlchemy: I’ve had scope issues before. Rarely but they do happen. Even if wasted bytes. This plugin gets like 5mil downloads a week. Eventually complaints would roll in. We are talking race conditions. 

Stupid question, what if we declared window as a variable, then w.document or w.location? One way to save bytes maybe. Can’t speak if race conditions would come of this never tried

jakearchibald: > We are talking race conditions.

I don't see a race condition here.

> what if we declared window as a variable, then w.document or w.location? One way to save bytes maybe.

If gzip's involved, that uses more bytes than `window.document`.

Fwiw I think omitting `window` is fine.

src/index.js
developit: Looks like Webpack core has also switched over to using `document.head`:
https://github.com/webpack/webpack/pull/8467

Kovensky:
Do we really need to be this concerned about older browser support if https://github.com/webpack-contrib/mini-css-extract-plugin/pull/299 is not included?

Any browser that this PR would break is already broken by not having #299.

evilebottnawi:
@Kovensky hm, I forgot that we need to use `webpack@4` for this plugin :smile: So `webpack@4` doesn't support IE8, right?

Kovensky:
It depends on whether it's still possible to make a bundle that does not _invoke_ `Object.defineProperties` (or `__webpack_require__.d` / `__webpack_require__.n` in this case; `__webpack_require_.r` can be faked but not `__webpack_require__.d` or `__webpack_require__.n`)

evilebottnawi:
@Kovensky thanks for information, what do you think should we merge this as patch or major release?

Kovensky:
It looks like it is still possible if your code and entire dependency graph consists **exclusively** of CommonJS or AMD modules... I find that extremely unlikely.

Personally I'm OK with this break as it's overwhelmingly unlikely that it will break bundles that weren't already broken, but I'm not sure if this might still affect people. It's not impossible that people might be intentionally using `babel-loader` + `@babel/plugin-transform-modules-commonjs` in `loose: true` mode to stop webpack from being able to see ES modules.

ScriptedAlchemy:
What about a feature release? Unless we are really making a heavy impact change? Are we losing compatibility with current PRd changes? 

Don’t mind major version bump, just throwing the question out there

ScriptedAlchemy:
Yeah that’s fine then. Lets get this in as a major

sokra:
People are using webpack 4 for ie8. That's a supported use case. ESM is not supported, by there is a ton of commonjs or amd code that can be used.

PR #336, opened 2019-01-09
feat: export class names as an ESM

### LINKED ISSUE #43: Export locals as ES2015 Modules
Hi,

Right now when I use this plugin my bundle ends up having a bunch of these:

```js
function(e, t, n) {
    e.exports = {
      home: "home_4zXj",
      headerContents: "headerContents_fgfT",
      info: "info_SpI-",
      dcmp: "dcmp_7LbX",
      cmp: "cmp_FYP2",
      pre: "pre_1JF_",
      off: "off_2kzf",
      adminPanel: "adminPanel_VjUM",
      navigationDrawerButtonContainer: "navigationDrawerButtonContainer_Gjyw",
      navigationDrawerButton: "navigationDrawerButton_11km"
    };
  }
```

Since this does not use es6 exports and instead uses `module.exports`, webpack is unable to inline these. If this plugin did use es6 exports, my bundle size would be much smaller because they could be inlined:

Code coming out of this plugin:
```js
export const home = "home_4zXj";
export const headerContents = "headerContents_fgfT";
export const info = "info_SpI-";
// etc.

export default {
  home: "home_4zXj",
  headerContents: "headerContents_fgfT",
  info: "info_SpI-",
  dcmp: "dcmp_7LbX",
  cmp: "cmp_FYP2",
  pre: "pre_1JF_",
  off: "off_2kzf",
  adminPanel: "adminPanel_VjUM",
  navigationDrawerButtonContainer: "navigationDrawerButtonContainer_Gjyw",
  navigationDrawerButton: "navigationDrawerButton_11km"
};
```

Then, after Webpack processes it, the css module would be inlined, and unused properties would be removed.

I made a fork of `extract-text-webpack-plugin` that does this: https://github.com/webpack-contrib/extract-text-webpack-plugin/compare/master...calebeby:master

Would you accept a similar PR to this repo?

Thanks!!!

que-etc:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Closes #43.

Consuming class names as an ESM with named exports makes Webpack to include them as separate bindings, which in turn, allows minifires to mangle them or inline those class names.

It all leads to a reduced size of a bundle, which, in case of our app, was `~30Kib` of a minified code.

### Breaking Changes

This a backward-compatible change due to an interoperability between ESM and CJS implemented in Webpack.

### Additional Info

Tests weren't updated because none of them assert exports of JS modules. Unfortunately, I've no idea how it can be tested given the existing cases.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -130,7 +130,13 @@ export function pitch(request) {
     }
     let resultSource = `// extracted by ${pluginName}`;
     if (locals && typeof resultSource !== 'undefined') {
-      resultSource += `\nmodule.exports = ${JSON.stringify(locals)};`;
+      Object.keys(locals).forEach((exportName) => {
+        const className = locals[exportName];
+
+        resultSource += `\nexport const ${exportName} = '${className}';`;
+      });
+
+      resultSource += `\nexport default ${JSON.stringify(locals)};`;
     }
 
     return callback(null, resultSource);


que-etc:
@evilebottnawi from an application's perspective or that it might break integration with other plugins?

evilebottnawi:
@que-etc 

> from an application's perspective

Also we need first implement this in `css-loader` and `style-loader`

src/loader.js
```diff
@@ -130,7 +130,13 @@ export function pitch(request) {
     }
     let resultSource = `// extracted by ${pluginName}`;
     if (locals && typeof resultSource !== 'undefined') {
-      resultSource += `\nmodule.exports = ${JSON.stringify(locals)};`;
+      Object.keys(locals).forEach((exportName) => {
+        const className = locals[exportName];
+
+        resultSource += `\nexport const ${exportName} = '${className}';`;
```
kgrz: Will this work with class names that have hyphens in them? Especially if `css-loader` isn't specifically configured to convert the locals into camelCase or something.

evilebottnawi:
Done in master

Issue #147, opened 2018-05-15
Breaking JS entrypoint when extracting all css chunks into 1 file


octaharon:
Package version: 0.4.0
Node version: 6.8.0

So I've updated to webpack 4 and instantly run into [extract-text-webpack-plugin incompatibility](https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/701), and guys on stackoverflow insisted on switching in favor of this module.  My use case is almost exactly [as described in documentation](https://github.com/webpack-contrib/mini-css-extract-plugin#extracting-all-css-in-a-single-file) with the difference being that there's sass-loader in the end of chain as well. All the rest is non-essential, the error is reproducable even with "zero" config. 

My project is React-based, with a set of components each having it's own scss file, required inside each component individually, including the Root component - the one which is rendered to DOM inside my entry file (say `/src/index.js`, has `require('index.scss')`).  

What happens on build stage is kinda in line with expectations - all the css got exported to `style.css` (despite it produces extra files like `style.main.js` for no obvious reason). The problem is that that my entrypoint javascript never got executed when the bundle is loaded. So there's no warnings or errors, everything compiles smoothly, the files (`main.js` and `style.css`) are delivered to the browser, but the code from `index.js` is not executed, so I just stare at a blank html template. 

In the end I switched back to `extract-text-webpack-plugin@next` cause I couldn't overcome that issue. Any ideas what causes that?

blakedietz:
I too was seeing an issue similar to this. Instead I was getting a third bundle created. For example if the bundle was named `bundle.js` I would see an output of `0.bundle.js, bundle.js, styles.css`.

evilebottnawi:
Feel free to investigate and send PR.
Don't use `extract-text-webpack-plugin@next` it is break long term cache, output very large bundle and can raise problem in runtime which very hard to find

octaharon:
Sorry, I've realized that a link to documentation was missing in my original post, edited. That's quite important for understanding the problem.

Perhaps the issue with plural output files belongs to webpack itself, since its driven by `optimization.splitChunks.cacheGroups`  config option, but the issue with "swallowing" the entrypoint of scripts is definitely coming from this plugin. If those are connected, than there's no better description 
>problems in runtime code which very hard to find and fix

[...]

PR #339, opened 2019-01-16
Prune empty JS bundles when plugin is used in conjunction with code splitting

rowanoulton:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

This is a proposed fix for #147. It's a stab in the dark and, while it seems to work, it may be the wrong approach entirely. Please advise!

For reference: the code samples below are from [this reproduction](https://github.com/dcousineau/mini-css-extract-plugin-reproduction) shared by @dcousineau.

#### Outline of the issue
When MiniCssExtractPlugin is used in conjunction with code splitting and the split chunk contains nothing other than CSS, webpack will still emit a JS file. Consider the following configuration:

```js
optimization: {
	splitChunks: {
		cacheGroups: {
		styles: {
			name: "styles",
			test: /\.css$/,
			chunks: "all",
			enforce: true
		}
	}
}
```

This will move all CSS across the entire compilation into a single file, `styles.bundle.css`. However, the compilation will also create a JavaScript bundle for this chunk under the same name: `styles.bundle.js`.

> **Aside:** I'd be curious to know why there's a JS module for every CSS module. My guess is the plugin needs a JS module to "hang" its CSS modules off of?

Often these bundles are more or less empty, save for some boilerplate:

```js
(window["webpackJsonp"] = window["webpackJsonp"] || []).push([["styles"],{

/***/ "./src/index.css":
/*!***********************!*\
  !*** ./src/index.css ***!
  \***********************/
/*! no static exports found */
/***/ (function(module, exports, __webpack_require__) {

// extracted by mini-css-extract-plugin

/***/ }),

/***/ "./src/sync.css":
/*!**********************!*\
  !*** ./src/sync.css ***!
  \**********************/
/*! no static exports found */
/***/ (function(module, exports, __webpack_require__) {

// extracted by mini-css-extract-plugin

/***/ })

}]);
```

This makes sense. The bodies of these modules are CSS, and the plugin has extracted them leaving only a comment behind.

#### Problematic bundles

These empty JavaScript bundles are problematic in that they must be loaded — and loaded in the correct order — for the application to boot properly. This is due to the way webpack boots, the logic of which is contained in webpack's boilerplate code. Here're the relevant parts:

```js
/******/ 	var jsonpArray = window["webpackJsonp"] = window["webpackJsonp"] || [];
/******/ 	var oldJsonpFunction = jsonpArray.push.bind(jsonpArray);
/******/ 	jsonpArray.push = webpackJsonpCallback;
/******/ 	jsonpArray = jsonpArray.slice();
/******/ 	for(var i = 0; i < jsonpArray.length; i++) webpackJsonpCallback(jsonpArray[i]);
/******/ 	var parentJsonpFunction = oldJsonpFunction;
/******/
/******/
/******/ 	// add entry module to deferred list
/******/ 	deferredModules.push(["./src/index.js","styles"]);
/******/ 	// run deferred modules when ready
/******/ 	return checkDeferredModules();
```

This code is responsible for installing modules: executing the function each module is contained in and returning its exports. Specifically, the install handler (`webpackJsonpCallback`) is executed for every module that was loaded before the main script.

Let's break it down. Here's a simplified version of`styles.bundle.js` again:

```js
window["webpackJsonp"] = window["webpackJsonp"] || []).push([["styles"], {
	(function(module, exports, __webpack_require__) {
		// extracted by mini-css-extract-plugin
	}),

	(function(module, exports, __webpack_require__) {
		// extracted by mini-css-extract-plugin
	})
}];
```

You can see that, as each file loads, it injects its modules into `window['webpackJsonp']`. 

When the main script finally runs, it iterates over these same modules and installs them via `webpackJsonpCallback`:

```js
/******/ 	var jsonpArray = window["webpackJsonp"] = window["webpackJsonp"] || [];
/******/ 	var oldJsonpFunction = jsonpArray.push.bind(jsonpArray);
/******/ 	jsonpArray.push = webpackJsonpCallback;
/******/ 	jsonpArray = jsonpArray.slice();
/******/ 	for(var i = 0; i < jsonpArray.length; i++) webpackJsonpCallback(jsonpArray[i]);
```

After all the modules are installed, webpack checks it has all the modules it needs (ie. "deferred modules"):

```js
/******/ 	// add entry module to deferred list
/******/ 	deferredModules.push(["./src/index.js","styles"]);
/******/ 	// run deferred modules when ready
/******/ 	return checkDeferredModules();
```

Note the mention of a module called "styles": this is that empty script we saw earlier, produced by the combination of MiniCssExtractPlugin and webpack's code splitting functionality.

Here's what webpack does when it checks deferred modules:

```js
/******/ 	function checkDeferredModules() {
/******/ 		var result;
/******/ 		for(var i = 0; i < deferredModules.length; i++) {
/******/ 			var deferredModule = deferredModules[i];
/******/ 			var fulfilled = true;
/******/ 			for(var j = 1; j < deferredModule.length; j++) {
/******/ 				var depId = deferredModule[j];
/******/ 				if(installedChunks[depId] !== 0) fulfilled = false;
/******/ 			}
/******/ 			if(fulfilled) {
/******/ 				deferredModules.splice(i--, 1);
/******/ 				result = __webpack_require__(__webpack_require__.s = deferredModule[0]);
/******/ 			}
/******/ 		}
/******/ 		return result;
/******/ 	}
```

It checks that every module in the array _after_ `./src/index.js` has been installed, or "fulfilled", before it does anything else.

This is critical: If the module "styles" has not been installed already, the boot will stall. This is because the `if(fulfilled)` branch doesn't execute, and it contains the first invocation of `__webpack_require__`, which is responsible for executing the main entry point and therefore the whole application.

#### Proposed fix

Here's what we want to do:

- Remove these empty JS files from the boot flow
- Keep our intended split chunks intact (the "styles" CSS bundle in our example)

And, if possible:

- Don't emit the empty JS files at all

#### Fixing the boot flow

The problem is what's written to webpack's boilerplate code:

```js
/******/ 	// add entry module to deferred list
/******/ 	deferredModules.push(["./src/index.js","styles"]);
```

We need to catch and prune empty files before they're included here.

Those exact lines are written out by the [JsonpMainTemplate](https://github.com/webpack/webpack/blob/master/lib/web/JsonpMainTemplatePlugin.js) plugin, invoked by [the creation of chunk assets](https://github.com/webpack/webpack/blob/0293c3a5ee6e41d04f4c5e1fb0607fdb7a0b38a8/lib/Compilation.js#L2386-L2393) during compilation. This is convenient as MiniCssExtractPlugin [already taps that event](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/18c066e58a0f29919df2c8f18861056f1464ce75/src/index.js#L166-L191):

```js
compilation.mainTemplate.hooks.renderManifest.tap(
	pluginName,
	(result, { chunk }) => {
	  const renderedModules = Array.from(chunk.modulesIterable).filter(
	    (module) => module.type === MODULE_TYPE
	  );
	  if (renderedModules.length > 0) {
	    result.push({
	      render: () =>
	        this.renderContentAsset(
	          compilation,
	          chunk,
	          renderedModules,
	        ),
	      filenameTemplate: this.options.filename,
	      pathOptions: {
	        chunk,
	        contentHashType: MODULE_TYPE,
	      },
	      identifier: `${pluginName}.${chunk.id}`,
	      hash: chunk.contentHash[MODULE_TYPE],
	    });
	  }
	}
);
```

This block of code is responsible for extracting the CSS from each module. Once that extraction has happened, it seems as if we're free to do whatever we like with the leftover module. This presents an opportune moment to search for split chunks that are purely CSS and perform some tidyup:

```js
const splitChunks = compilation.chunks.filter(chunk => {
  return (
    chunk.chunkReason && chunk.chunkReason.includes("split chunk")
  );
});

splitChunks.forEach((splitChunk) => {
  const modulesWeCouldRemove = [];
  const nonCssModules = Array.from(splitChunk.modulesIterable).filter((mod) => {
    return mod.type !== MODULE_TYPE;
  });

  nonCssModules.forEach((nonCssMod) => {
    if (
      nonCssMod._source &&
      nonCssMod._source._value === `// extracted by ${pluginName}`
    ) {
      modulesWeCouldRemove.push(nonCssMod);
    }
  });

  // If there's nothing but CSS modules left in this split chunk, remove the whole thing
  // This will mean that the main template manifest (ie. webpack's boilerplate code) won't
  // contain any references to these empty modules
  if (modulesWeCouldRemove.length === nonCssModules.length) {
    modulesWeCouldRemove.forEach((nonCssMod) => {
      // Trace all the "reasons" for this module (ie. other modules that depend on it)
      // then add this module into their respective chunks. This effectively reverses
      // the work SplitChunksPlugin did to break out the logic into a separate chunk.
      // Without this step there will be script errors owing to missing dependencies
      // and adding them back to their origin chunks is harmless as they're empty.
      nonCssMod.reasons.forEach((reason) => {
        reason.module.chunksIterable.forEach((previouslyConnectedChunk) => {
          splitChunk.moveModule(nonCssMod, previouslyConnectedChunk);
        });
      });
    });

    splitChunk.groupsIterable.forEach((group) => {
      group.removeChunk(splitChunk);
    });
  }
});
```

For any purely CSS-based split chunks we reverse the work SplitChunksPlugin did by moving the empty JS modules back to their origin chunks then remove the split chunk from the module graph altogether. By this method we prevent them from being referenced by the boilerplate boot code.

#### Keeping the stylesheets in play

With the first problem solved, we encounter another: by removing empty split chunks from the dependency graph, the stylesheets are emitted by webpack but are ignored by plugins like `HtmlWebpackPlugin`. In practice this can mean HTML rendered without the appropriate `<link>` tags.

To combat this, we have to retain a dependency on our CSS split chunks. If we track which groups a given chunk was removed from, we can restore the connections after the boilerplate has been generated. First this requires a bit of extra book-keeping:

```js
class MiniCssExtractPlugin {
  constructor(options) {
    this.disconnectedGroups = {};
    ...
```

```js
splitChunk.groupsIterable.forEach((group) => {
  group.removeChunk(splitChunk);

  // Book-keeping
  this.disconnectedGroups[splitChunk.id] =
    this.disconnectedGroups[splitChunk.id] || [];
  this.disconnectedGroups[splitChunk.id].push(group);
});
```

The first event after the boilerplate code has been generated is `compilation.hooks.chunkAsset`, so we do our reconnection work there:

```js
compilation.hooks.chunkAsset.tap(pluginName, (chunk, file) => {
  // If this was a split chunk we disconnected previously
  if (this.disconnectedGroups[chunk.id]) {
    if (path.extname(file) === '.css') {
      this.disconnectedGroups[chunk.id].forEach((group) => {
        // Add the stylesheet as a file on every chunk that requires it
        // This ensures plugins like html-webpack-plugin will find and include
        // split CSS chunks
        group.chunks.forEach((parentChunk) => {
          parentChunk.files.push(file);
        });
      });
    }
  }
});
```

I'm especially unsure about this code but the intention is to reinsert the CSS files into every chunk that depends on those styles, and this appears to be enough to appease `HtmlWebpackPlugin`.

#### Discarding the empty JS files

From here it's just a small jump to discard the empty JS bundles altogether. We simply branch from the previous file extension check and, if it's JavaScript, delete it:

```js
if (path.extname(file) === '.css') {
  this.disconnectedGroups[chunk.id].forEach((group) => {
    // Add the stylesheet as a file on every chunk that requires it
    // This ensures plugins like html-webpack-plugin will find and include
    // split CSS chunks
    group.chunks.forEach((parentChunk) => {
      parentChunk.files.push(file);
    });
  });
} else if (path.extname(file) === '.js') {
  // Discard empty JS file
  chunk.files.pop();
  delete compilation.assets[file];
}
```

This is undoing the work that would have [just happened in the webpack compiler](https://github.com/webpack/webpack/blob/master/lib/Compilation.js#L2410-L2411).

#### Outcome

Applying these changes to the [aforementioned reproduction](https://github.com/dcousineau/mini-css-extract-plugin-reproduction) gives us the result we want:

- One CSS bundles (`styles.css`)
- No empty JS bundle (`styles.js`)
- An application that boots

I've confirmed the same is true of a webpack-dev-server build, and tested with both `html-webpack-plugin@latest` and `html-webpack-plugin@next`.

### Breaking Changes

Unclear. This change will mean empty JS bundles are no longer output for every CSS split chunk. That may break workarounds people have been using.

### Additional information

I'm not convinced this is the best way to fix this problem, and could use some input from someone better versed in this plugin and with webpack in general.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,5 @@
+import path from 'path';
+
 import webpack from 'webpack';
 import sources from 'webpack-sources';
 
@@ -110,6 +112,7 @@ class CssModuleFactory {
 
 class MiniCssExtractPlugin {
   constructor(options) {
+    this.disconnectedGroups = {};
     this.options = Object.assign(
       {
         filename: '[name].css',
@@ -188,8 +191,81 @@ class MiniCssExtractPlugin {
               hash: chunk.contentHash[MODULE_TYPE],
             });
           }
+
+          const splitChunks = compilation.chunks.filter(
+            (thisChunk) =>
+              thisChunk.chunkReason &&
+              thisChunk.chunkReason.includes('split chunk')
+          );
+
+          splitChunks.forEach((splitChunk) => {
+            const modulesWeCouldRemove = [];
+            const nonCssModules = Array.from(splitChunk.modulesIterable).filter(
+              (mod) => mod.type !== MODULE_TYPE
+            );
+
+            nonCssModules.forEach((nonCssMod) => {
+              if (
+                nonCssMod._source && // eslint-disable-line no-underscore-dangle
+                nonCssMod._source._value === `// extracted by ${pluginName}` // eslint-disable-line no-underscore-dangle
+              ) {
+                modulesWeCouldRemove.push(nonCssMod);
+              }
+            });
+
+            // If there's nothing but CSS modules left in this split chunk, remove the whole thing
+            // This will mean that the main template manifest (ie. webpack's boilerplate code) won't
+            // contain any references to these empty modules
+            if (modulesWeCouldRemove.length === nonCssModules.length) {
+              modulesWeCouldRemove.forEach((nonCssMod) => {
+                // Trace all the "reasons" for this module (ie. other modules that depend on it)
+                // then add this module into their respective chunks. This effectively reverses
+                // the work SplitChunksPlugin did to break out the logic into a separate chunk.
+                // Without this step there will be script errors owing to missing dependencies
+                // and adding them back to their origin chunks is harmless as they're empty.
+                nonCssMod.reasons.forEach((reason) => {
+                  reason.module.chunksIterable.forEach(
+                    (previouslyConnectedChunk) => {
+                      splitChunk.moveModule(
+                        nonCssMod,
+                        previouslyConnectedChunk
+                      );
+                    }
+                  );
+                });
+              });
+
+              splitChunk.groupsIterable.forEach((group) => {
+                group.removeChunk(splitChunk);
+
+                // Book-keeping
+                this.disconnectedGroups[splitChunk.id] =
+                  this.disconnectedGroups[splitChunk.id] || [];
+                this.disconnectedGroups[splitChunk.id].push(group);
+              });
+            }
+          });
         }
       );
+      compilation.hooks.chunkAsset.tap(pluginName, (chunk, file) => {
+        // If this was a split chunk we disconnected previously
+        if (this.disconnectedGroups[chunk.id]) {
+          if (path.extname(file) === '.css') {
+            this.disconnectedGroups[chunk.id].forEach((group) => {
+              // Add the stylesheet as a file on every chunk that requires it
+              // This ensures plugins like html-webpack-plugin will find and include
+              // split CSS chunks
+              group.chunks.forEach((parentChunk) => {
+                parentChunk.files.push(file);
+              });
+            });
+          } else if (path.extname(file) === '.js') {
+            // Discard empty JS file
+            chunk.files.pop();
+            delete compilation.assets[file]; // eslint-disable-line no-param-reassign
+          }
+        }
+      });
       compilation.chunkTemplate.hooks.renderManifest.tap(
         pluginName,
         (result, { chunk }) => {


rowanoulton:
@evilebottnawi thanks for the quick reply. Do you have some thoughts on how this should be fixed in webpack?

RE: CI failures. Are they worth fixing if this isn't the right approach? The failure is in regard to `npm audit` which I'm unsure is related to my change: 

```
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ High          │ Missing Origin Validation                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ webpack-dev-server                                           │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ webpack-dev-server [dev]                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ webpack-dev-server                                           │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/725                       │
└───────────────┴──────────────────────────────────────────────────────────────┘
```

evilebottnawi:
@rowanoulton webpack doesn't should emit empty chunks, it is should be fixed in `webpack@5` in split chunks

evilebottnawi:
Update lock file and commit

Jab2870:
I am still getting this issue. Has it been fixed? Is there a workaround?

rowanoulton:
@evilebottnawi I took [your earlier comment](https://github.com/webpack-contrib/mini-css-extract-plugin/pull/339#issuecomment-455098461) to mean this isn't the right approach. The main issue to patch being [this one](https://github.com/webpack/webpack/issues/7300#issuecomment-495599502). Is that right? if so, I'll close this PR.

rowanoulton:
For those following along, it looks like [this](https://github.com/webpack/webpack/pull/9040) is the fix.

rowanoulton:
Relevant commit [here](https://github.com/webpack/webpack/commit/c5f94f3b6a79a88da9ed93b5f980830f496f4fad)

PR #345, opened 2019-02-01
fix: onerror  request of event.target should be href rather than src

WilsonLiu95:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

The url of link should be href rather than src. Although now  has fullhref final return.
```
'var request = event && event.target && event.target.src || fullhref;',
```


WilsonLiu95:
> Why? Can you describe?

```
'var request = event && event.target && event.target.src || fullhref;',
```

Issue #204, opened 2018-07-06
Async loading of CSS bundle


Vangor:
Is there a way to disable  async loading of CSS bundle?
Using it with this config.

```
new MiniCssExtractPlugin({
  chunkFilename: 'data-police.css?t=[contenthash]'
})
```

```
MiniCssExtractPlugin.loader,
{
  loader: 'css-loader',
  options: {
  importLoaders: 1,
  sourceMap: true,
  url: false,
  modules: true,
  localIdentName: '[path]___[name]__[local]___[hash:base64:5]'
  }
 }
```

evilebottnawi:
@Vangor What you mean? Just don't load css async

Vangor:
@evilebottnawi Well, the thing is, that I don't specify it anywhere. I just use these settings, and CSS file is loaded asynchronously. 
Maybe it's how this plugin works with css-modules?

Vangor:
I will try but don't think so.

evilebottnawi:
@Vangor Without reproducible test repo, we can't help, sorry

zahorovskyi:
@evilebottnawi Hi. I have small question regarding to your comment "@Vangor What you mean? Just don't load css async". Do we have some way to disable just async loading?

ScriptedAlchemy:
i get what you want. 

you are looking for something that will extract the stylesheets, but not necessarily append them on demand when a new javascript file is consumed. 

From my development of a derivative of mini-css, i can tell you that its pretty baked into the plugin, as it should be. 

The feature you are requesting does not really align with the goals of webpack-team. 

Mini css likely will be a built in part of webpack in 5.x 

if you want to remove the ability to load on demand (append a `link` to the head), you might want to either look for another loader which may further process the file, or consider changing your applications architecture so that it can work with webpack.

There some things to consider as well, like chunking, and the fact that mini-css actually does a good job at optimizing and combining chunks... The file you want, might have been transformed and shares styles with other components / relys on other style optimizations and its own loading mechanism to recombine the used styles. 

One workaround might be to just use file-loader and postcss, it wouldn't have the loading mechanism to append resources on demand like mini-css does.

Vangor:
Thanks for Explanation @ScriptedAlchemy 
I think we will use it as is for now.
This was needed, since we have microservice architecture, and controller that brings all services together, loads it's css and js, and we wanted to handle it trough controller only.

ScriptedAlchemy:
[...]

PR #348, opened 2019-02-12
Added option to disable async loading of CSS

thomasbaldwin:
This PR contains a:

- [ ] **bugfix**
- [X] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

The reason for this PR are for the scenarios in which you have multiple branded websites, all using the same components, but different styles based off the brand. Lets say we have a component that looks like this:

```
import React from 'react';
import PropTypes from 'prop-types';
import { A, B, C } from 'config/brands';

const brandStyles = {};
brandStyles[A] = require('./styles.a.scss'); // version A of the component styling
brandStyles[B] = require('./styles.b.scss'); // version B of the component styling
brandStyles[C] = require('./styles.c.scss'); // version C of the component styling

const BrandedComponent = ({ brand, ... }) => {
  const styles = brandStyles[brand] || brandStyles[A];

  return (
    <div className={styles.container}>
      ...
    </div>
  )
}

export default BrandedComponent;
```

Notice how there's an A, B, and C stylesheet. This is because using a different stylesheet, we can drastically change how the component looks in our application, essentially creating different branded versions of it. Depending on the brand you hit, the correct stylesheet will be applied.

The way I've compiled the stylesheets is so that there's a `global.css`, `a.css` (all the styles.a.scss files), `b.css` (all the styles.b.scss files), `c.css` (all the styles.c.scss files) ...etc files. Depending on the brand you hit, I attach the correct stylesheets in a link tag to the page source so the browser will download them.

If I had not disabled the async loading of styles, the other brand styles being required at the top of the component would have been asynchronously downloaded, sometimes causing the component to not look as I wanted it to.

This has been asked for before in issues, so I figured we can't be the only ones who needed a feature like this:

https://github.com/webpack-contrib/mini-css-extract-plugin/issues/204

With this PR, you can now disable mini css extract from asynchronously downloading other stylesheets, by providing it the `disableAsync` option.

```
new MiniCssExtractPlugin({ filename: "[name].css", disableAsync: true })
```

### Breaking Changes

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,6 +367,16 @@ new MiniCssExtractPlugin({
 }),
 ```
 
+### Disable Asynchronous Downloading of CSS
+
+For projects in which you do not want CSS asynchronously fetched on the client side, and only want the CSS that is delivered on the server side (via link tags) to be fetched and parsed, enabling this option will give you that functionality.
+
+```javascript
+new MiniCssExtractPlugin({
+  disableAsync: true,
+}),
+```
+
 ### Media Query Plugin
 
 If you'd like to extract the media queries from the extracted CSS (so mobile users don't need to load desktop or tablet specific CSS anymore) you should use one of the following plugins:
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "test:watch": "cross-env NODE_ENV=test jest --watch",
     "test:coverage": "cross-env NODE_ENV=test jest --collectCoverageFrom=\"src/**/*.js\" --coverage",
     "test:manual": "webpack-dev-server test/manual/src/index.js --open --config test/manual/webpack.config.js",
+    "test:manualDisableAsync": "webpack-dev-server test/manualDisableAsyncOption/src/index.js --open --config test/manualDisableAsyncOption/webpack.config.js",
     "pretest": "npm run lint",
     "test": "cross-env NODE_ENV=test npm run test:coverage",
     "defaults": "webpack-defaults"
--- a/src/index.js
+++ b/src/index.js
@@ -304,6 +304,8 @@ class MiniCssExtractPlugin {
       mainTemplate.hooks.requireEnsure.tap(
         pluginName,
         (source, chunk, hash) => {
+          if (this.options.disableAsync) return null;
+          
           const chunkMap = this.getCssChunkObject(chunk);
 
           if (Object.keys(chunkMap).length > 0) {
--- a/src/index.js
+++ b/src/index.js
@@ -305,7 +305,6 @@ class MiniCssExtractPlugin {
         pluginName,
         (source, chunk, hash) => {
           if (this.options.disableAsync) return null;
-          
           const chunkMap = this.getCssChunkObject(chunk);
 
           if (Object.keys(chunkMap).length > 0) {
--- a/test/manualDisableAsyncOption/index.html
+++ b/test/manualDisableAsyncOption/index.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8" />
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <title>mini-css-extract-plugin testcase</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <link rel="stylesheet" type="text/css" href="/dist/a.css" />
+</head>
+<body>
+  <script type="text/javascript" src="/dist/a.js"></script>
+  <script type="text/javascript" src="/dist/b.js"></script>
+  <script type="text/javascript" src="/dist/main.js"></script>
+</body>
+</html>
--- a/test/manualDisableAsyncOption/src/a.css
+++ b/test/manualDisableAsyncOption/src/a.css
@@ -1,0 +1,7 @@
+.container {
+  background: blue;
+  color: red;
+  width: 100%;
+  height: 20px;
+  text-align: center;
+}
--- a/test/manualDisableAsyncOption/src/b.css
+++ b/test/manualDisableAsyncOption/src/b.css
@@ -1,0 +1,7 @@
+.container {
+  background: purple;
+  color: green;
+  width: 100%;
+  height: 20px;
+  text-align: center;
+}
--- a/test/manualDisableAsyncOption/src/index.js
+++ b/test/manualDisableAsyncOption/src/index.js
@@ -1,0 +1,14 @@
+/* eslint-env browser */
+
+import './a.css';
+import './b.css';
+
+const render = () => {
+  const div = document.createElement('div');
+  div.setAttribute('class', 'container');
+  const text = 'My background color should be blue, and my text should be red!';
+  div.innerHTML = text;
+  document.body.appendChild(div);
+};
+
+render();
--- a/test/manualDisableAsyncOption/webpack.config.js
+++ b/test/manualDisableAsyncOption/webpack.config.js
@@ -1,0 +1,45 @@
+const Self = require('../../');
+
+module.exports = {
+  mode: 'development',
+  output: {
+    chunkFilename: '[name].js',
+    publicPath: '/dist/',
+    crossOriginLoading: 'anonymous',
+  },
+  optimization: {
+    splitChunks: {
+      cacheGroups: {
+        default: false,
+        a: {
+          name: 'a',
+          test: /a.css/,
+          chunks: 'all',
+          enforce: true,
+        },
+        b: {
+          name: 'b',
+          test: /b.css/,
+          chunks: 'all',
+          enforce: true,
+        },
+      },
+    },
+  },
+  module: {
+    rules: [
+      {
+        test: /\.css$/,
+        use: [Self.loader, 'css-loader'],
+      },
+    ],
+  },
+  plugins: [
+    new Self({
+      filename: '[name].css',
+    }),
+  ],
+  devServer: {
+    contentBase: __dirname,
+  },
+};


evilebottnawi:
Why don't use `const brandStyles = {}; if (something) { brandStyles[A] = require('./styles.a.scss'); }`?

thomasbaldwin:
Hey @evilebottnawi, thanks for your reply! I assume you're talking about something like this:

```
import React from 'react';
import PropTypes from 'prop-types';
import { A, B, C } from 'config/brands';

const { brand } = process.env;

const brandStyles = {};
if (brand === A) brandStyles[A] = require('./styles.a.scss'); // version A of the component styling
if (brand === B) brandStyles[B] = require('./styles.b.scss'); // version B of the component styling
if (brand === C) brandStyles[C] = require('./styles.c.scss'); // version C of the component styling

const BrandedComponent = ({ brand, ... }) => {
  const styles = brandStyles[brand];

  return (
    <div className={styles.container}>
      ...
    </div>
  )
}

export default BrandedComponent;
```

The reason we can't do something like this is because this would require the application to be built for a specific brand at a time (notice the brand condition is coming from an environment variable). By passing in the brand to be served through props, this gives you the ability to serve all brands in the same server and the decision of which brand to load is made during runtime.

evilebottnawi:
@thomasbaldwin you can use `document.querySelector('#theme')` instead `process.env`. You try to solve problem what should be solved on code side. Also disabled async increase you bundle size, it is really bad for you clients (why i need download style of theme which i never will be used?) and bad option for plugin.

thomasbaldwin:
@evilebottnawi document.querySelector is not isomorphic... The bundle sizes are not any bigger because we have all the different branded CSS put into their own compiled stylesheets (i.e. brandA.css, brandB.css, brandC.css) Based off the brand you hit, we attach the correct stylesheet to the page (so only the styles that are being used are being downloaded). We were able to cut the CSS we're delivering for each brand in half following this approach.

The only problem is when you load a component, mini css extract would asynchronously download the other brands stylesheets that we did not attach to the page source. The disableAsync option is there to make it so it won't do that if you don't want it to

evilebottnawi:
@thomasbaldwin 

> document.querySelector is not isomorphic

What do you mean?

> The only problem is when you load a component, mini css extract would asynchronously download the other brands stylesheets that we did not attach to the page source. The disableAsync option is there to make it so it won't do that if you don't want it to

It is break logic for async concept, just don't use plugin and include css using `raw-loader` or include their in `style`/`link` tag

thomasbaldwin:
@evilebottnawi what I mean by `document.querySelector` not being isomorphic is that I can't say `document.querySelector` on the server side of my application because there doesn't exist a document. Only on the client side can I use a command like that. Our application is universal and the decision of which brand to load needs to be made on the server side.

The reason I've chosen mini css extract is because it will write the CSS to files (which I don't believe raw-loader will do), and using the splitChunks optimization, I'm able to configure which files are generated and how.

This is perfect, however, the issue we've run into is that when we attach a specific brand styling to the page source (i.e. `<link href="brandA.css" rel="stylesheet" />`), mini css extract would see the other requires in my component and go fetch them, which is why I believe there is a need for something like this.

evilebottnawi:
@thomasbaldwin Why don't use `document && document.querySelector ? document.querySelector('#theme') : defaultTheme`? I think it is normal use default theme when client don't choose you preferred theme. 

You can save `theme` in cookie and use `document && document.querySelector ? document.querySelector('#theme') : cookie.get('theme') ? cookie.get('theme') : defaultTheme`

I don't think we really need this option, because right way is solve this problem on code side. It can also create bad practices for other developers.

thomasbaldwin:
@evilebottnawi that’s an interesting idea, but unfortunately the decision of which brand to load is not made by the client.

It’s actually made by the CMS’ API. We give it the URL that’s been hit, and it tells us what brand the page is and what components should appear on that page. Essentially every page is completely dynamic and controlled by the CMS.

I really like mini css extracts ability to create multiple compiled CSS files, I just wish we had some control over it asynchronously downloading things

evilebottnawi:
@thomasbaldwin 

> We give it the URL

You need `js` API for this on CMS side. All works with CMS features should be done using API of CMS.

I wait answer from @sokra , but i vote :-1: on this option, it is bad practice and some new developers can starts using this.

dmarcs:
@thomasbaldwin +1 I could really use something like this

evilebottnawi:
@dmarcs what is you use case? You really want to use bad practice?

dmarcs:
@evilebottnawi - I also would like to have the ability to create multiple compiled stylesheets rather than just 1 (which is what extract text plugin forces us to do). However, that is all I need out of mini css extract. I don’t need it to go and download the other styles I haven’t manually referenced via a link tag.

I don’t think this is a bad practice by any means. For some of us, mini css extract is doing too much and downloading stylesheets being referenced when we would like it not to.

evilebottnawi:
@dmarcs 

> @evilebottnawi - I also would like to have the ability to create multiple compiled stylesheets rather than just 1 (which is what extract text plugin forces us to do). However, that is all I need out of mini css extract. I don’t need it to go and download the other styles I haven’t manually referenced via a link tag.

Use `splitChunks` for this. This option doesn't solve your problem anyway.

dmarcs:
@evilebottnawi - how do I compile the stylesheets and write them to files without mini css extract? I am planning on using split chunks to split the CSS up, but what will turn my sass into css and write it to a file besides mini css extract? 

What I don’t need is mini css extract to go and download CSS just because it sees it was required. I would like to have control over that. This feature would solve a lot of my problems

evilebottnawi:
@dmarcs 

> I am planning on using split chunks to split the CSS up, but what will turn my sass into css and write it to a file besides mini css extract?

Please read documentation how `splitChunks` works and try examples.

> This feature would solve a lot of my problems

No, it is create new problems for you and your clients.

dmarcs:
@evilebottnawi - as @thomasbaldwin pointed out, there are times when you want to have a skinned version of component driven by different stylesheets. Using split chunks and mini css extract, we can easily create stylesheets that contain only what is necessary for certain skinned versions to run. The issue though is that mini css extract will go and download styles being required, even if the runtime logic in the render method doesn’t say to use that stylesheet. By having the ability to disable mini css extracts async downloading of stylesheets, this would give me full control over what styles are being loaded

evilebottnawi:
@dmarcs Please provide minimum reproducible test repo where this option required and you can't solve problem without this option

evilebottnawi:
@D-Marc1 It can't be merged without arguments, please provide your use case

D-Marc1:
Very similar issue to @thomasbaldwin. I have components that are controlled by different stylesheets, and am creating variation specific stylesheets:

For example:

Component1
  -variationA.css
  -variationB.css
  -variationC.css

Component2
  -variationA.css
  -variationB.css
  -variationC.css

Component3
  -variationA.css
  -variationB.css
  -variationC.css

Depending on the domain the user hits, I will render the same site just styled differently (with more or less components placed on different areas of the page). For example, say you hit variationA.com, you will be given the variationA.css stylesheet, which contains all the styles from the variationA.css files. Likewise, if you hit variationB.com, you will be given the variationB.css stylesheet, which contains all the styles from variationB.css files.

I have code split off the components, and I only attach the components necessary for the site to run. However, because my components require all the different variation stylesheets in the global space, the other stylesheets are being fetched which results in more CSS than I need and also affects the styling of the component.

I know many people that are doing similar architectures to this in which they serve multiple branded websites off a single platofrm / server, and the decision of which brand / stylesheet to serve is made during runtime.

Having the ability to disable the async loading of stylesheets should have been an option from the beginning in my opinion. Thanks @evilebottnawi

evilebottnawi:
@D-Marc1 you should load styles based on `domain` in your url, async logic same for any assets, like js, css, markdown and etc.

evilebottnawi:
It is not extract plugin, it is plugin for css. Extract is a part of this job.

thomasbaldwin:
@evilebottnawi - What if there is no domain difference between variations? For example, you could have variations1 being served here:

**website.com/this-is-url7138901312**
**website.com/this-is-another-url3**

and variations2 being served here:

****website.com/this-a-url-2**
**website.com/this-is-another-url712612****

Because there's not an entire pattern I can rely on, I have to rely on my API rather than the domain to tell me what variation the user is hitting

evilebottnawi:
@thomasbaldwin You should use CMS/framework API to get theme value, as i said above. All what you want should be done on `code` side, and most likely this option will never be merged.

evilebottnawi:
Just note: PR should contains tests

thomasbaldwin:
@evilebottnawi sounds good I will add tests. I will also make a minimum reproducible repo so you can see the issue I'm running into and why I needed this feature. Thanks for your help :)

lsycxyj:
@evilebottnawi I don't know if I've misunderstood the purpose of this pull request. If it's used to disable extracting the css of all async chunks, it think it should be useful. Because I don't need to make extra an request to fetch the css file of its own async js chunk which it has been loaded! It should be something like the "fallback" option of `extract-text-webpack-plugin` and its behavior.

For example, let's say we have several pages will be loaded on demand:
```javascript
// routes.js
const routes: [
  {
    module: () => import('pageA')
  }
];

// pageA.js
import './pageA.css';
export default {
  // something that pageA should do
};
```
And the output file should be like this:
```
vendor.js
initial-chunk.js
initial-chunk.css
pageA.js
// Note that there isn't a "pageA.css" file
```
I wonder why the "fallback" option was removed in this plugin.

rseeberg:
This would be a very helpful thing to have.
We are server the build output from a asp.net application thats including all of the dist files.
Due to miniCssExtractPlugin doing some async loading magic at runtime, we are getting some css files duplicated.
This is incredibly annoying. It would be very much appreciated to have a "EXTRACT" plugin that does just that, extract, and not worries about loading. I'm perfectly capable of serving my css files as I see fit.

evilebottnawi:
Please rebase

rjgotten:
> @dmarcs 
> The issue though is that mini css extract will go and download styles being required, even if the runtime logic in the render method doesn’t say to use that stylesheet.

You can easily work around that by using context modules.
The trick starts with something like this in your JS:

```js
import( /* webpackMode: lazy*/ `./component.${tenant}.scss`);
```

where `tenant` can come from any dynamic runtime source you want.

And then the secret sauce is to put a cache group configuration like _this_ in your webpack config:

```js
cacheGroups : {
  tenants : {
    test : module => {
      const resource = module.constructor.name === "CssModule"
        ? module.issuer.resource
        : module.resource;
    
      return /\.[^\s.]+\.scss$/.test( resource );
    },
    name : module => {
      const resource = module.constructor.name === "CssModule"
        ? module.issuer.resource
        : module.resource;

      const match = /\.([^\s.]+)\.scss$/.test( resource );
      const tenant = ( match && match[ 1 ]) || "default";
  
      return `styles/tenants/${tenant}`;
    },
    chunks  : "async",
    enforce : true
  }
}
```

Most don't know or realize, but the names for cache groups can be fed by a function and can be established _dynamically_ like this, meaning it can pull all files related to any single tenant together, without ever having to configure an explicit list of either tenants or files.

As MiniCssExtract honors cache groups and can create one-per-group CSS output, this gives you one-per-tenant CSS output sheets. And with the context module set as lazy, it's ensured that _only_ the one tenant requested at runtime will have its module actually loaded and thus its accompanying stylesheet hooked in.

----

Finally, there may indeed be good reasons to avoid having a dynamic load happening at all in production.
E.g. because you're only loading one big sheet up front per tenant and you only need the async loaded sheet in development mode, because it adds HMR support.

In those cases you can mask out the import using a `process.env` based guard that skips them in production mode:

```js
if ( process.env.ENV === "development" ) {
 import( /* webpackMode: lazy*/ `./component.${tenant}.scss`);
} 
```

rjgotten:
> @rseeberg 
> we are getting some css files duplicated.

As long as the path you put into your static `<link>` tag exactly matches the path used by MiniCssExtract when async loading the style, it will overwrite rather than duplicate.

This is actually great for hot module reloading.


The easiest way to ensure the paths match, is to make the Webpack public path domain-relative and to then have your ASP.NET MVC application consisntely output its URLs using app-relative paths with a leading `~`, which that framework will translate into domain-relative paths.

lsycxyj:
@evilebottnawi I made another PR #432 based on recent master branch.

thomasbaldwin:
@lsycxyj I will rebase and add some tests here. I took a look at your PR, it looks like you're changing a lot more than needs to be

ScriptedAlchemy:
So you want to disable dynamic `import()` loading the stylesheets subsequent stylesheets?

thomasbaldwin:
@ScriptedAlchemy - I want to disable mini css extract from fetching CSS that is even mentioned in normal "requires" in a component. Currently, if you list multiple CSS files using require in a JS file, and only attach one of those CSS files through a link tag, mini css extract will by default go and download the other css files on the client side

thomasbaldwin:
@evilebottnawi I would like to write a test for these changes, but I don't see any tests written that test the client side downloading of CSS files used in a require. I see tests around the build output but not around the runtime functionality

thomasbaldwin:
I've added in tests too @evilebottnawi

lsycxyj:
@thomasbaldwin I changed much more than your code because my feature can disable by each either entry or async module, instead of disabling all async modules.
In addition, your test case doesn't seem to be an async import. Or maybe my cases are completely different from yours.

anilanar:
I'm not convinced. @thomasbaldwin 

Your component somehow has access to current `brand`. So you know the `brand` during runtime.

You also know how to map a `brand` to which file. So you can easily do the following:

```
const getBrandedStyles = brand => import(
  /* webpackChunkName: "styles-[brand]" */
  `./styles.${brand}.css`
);

const App = ({ brand }) => {
  const [styles, setStyles] = useState(null);
  useEffect(() => {
    getBrandedStyles(brand).then(styles => setStyles(styles));
  }, [brand]);
  return (styles ? <Loading /> : <BrandedComponent styles={styles} />);
}
```

This will create a css file for every file matching `styles.___.css` pattern and your runtime will only download it based on `brand` value.

You can use `ContextReplacementPlugin` to do more advanced mapping between `brand` and css file name.

thomasbaldwin:
@anilanar how would this work with SSR? I’ll take a look at the plugin

anilanar:
@thomasbaldwin You implement a hook `useStyles(brand)`. `useStyles` has different impls based on environment. If the environment is node, you don’t use `useEffect`. It’d directly require the style (use babel-plugin for replacing dynamic imports with require etc.)

thomasbaldwin:
@anilinar - I’m assuming I will have to traverse through the component tree and call this hook on every component during SSR, keep track of the stylesheets they are trying to load, and attach the corresponding stylesheets to the page source. Is that a correct assumption?

anilanar:
@thomasbaldwin If you are trying to do what other CSS-in-JS solutions are already doing with their SSR, I’d check how they handle things before reinventing the wheel.

E.g instead of iterating the component tree, you can use a provider of your own (context api of react) and the `useStyles` hook would use the provider to request a style. This way, your provider can exactly know which styles are used during the rendering.

Rolling your own solution to this complex domain is not easy though.

thomasbaldwin:
@anilanar Do you have an example of a CSS-in-JS solution that will handle code splitting multiple CSS files per 1 JS file. I believe MiniCSSExtract will only let me code split 1 CSS file per JS

anilanar:
@thomasbaldwin CSS-in-JS is just Javascript files that represent styles with JS modules. So code splitting js and styles has no difference. You can handle your requirements you mention above e.g with styled-components and `ThemeProvider`: https://www.styled-components.com/docs/advanced

However going with css-in-js, it’s not as easy to apply styles to your whole DOM tree with just 1 css file. Because css-in-js is supposed to isolate styles at component level. That’s its power, and maybe in your case its weakness.

clucaseh:
Can we merge this?

anilanar:
@clucaseh Potentially yes, but I have two proposals/additions:

1) Maybe `disableAsync` could be renamed. What this option does is it prevents auto injection of `<link href>` elements to `document.head` when an async chunk is loaded via `require.ensure` or async `import`, similar to how extracted CSS are handled for entry chunks.
2) For entry CSS files, a webpack user either has to use HTML webpack plugin to auto generate `<link>` tags or generate a webpack manifest which would be consumed by the server that serves the actual page etc. For async chunks, when this option is enabled, css must be manually injected, but how? How does runtime code know the css file path for the current chunk? e.g:

    ```
    import('./a.js').then(_module => /* what to do here? */);
    ```

    One option is to solve it at loader level. `mini-css-extract-plugin`'s loader would export the url of the css file:

    ```
    // a.js
    import url from 'a.css';
    import myCache from './my-css-cache';
    
    if (!myCache[url]) {
        document.head.appendChild(/* element for <link rel="stylesheet" href={url} /> */);
        myCache[url] = true;
    }
    ```

lsycxyj:
> @anilanar Do you have an example of a CSS-in-JS solution that will handle code splitting multiple CSS files per 1 JS file. I believe MiniCSSExtract will only let me code split 1 CSS file per JS file

Can [this PR](https://github.com/webpack-contrib/mini-css-extract-plugin/pull/432) solve your problem?

clucaseh:
> 2\. e option is to solve it at loader level. `mini-css-extract-plugin`'s loader would export the url of the css file:
>     ```
>     // a.js
>     ```

That sounds doable. @thomasbaldwin do you want to work on this?

I currently don't use html webpack plugin. We use c# razor so we don't use html pages. We are loading our main.css file in our layout as we want it to load as fast as possible. Currently trying to find a way to generate the CSS file but not have it imported in the final bundle. 

Only way I have found to do this is too run two scripts on our end. First time we generate both the CSS and JS. Run script again passing null-loader for CSS/Sass so no CSS is generated in the final JS. Which, will work it will just increase our build time. So this option would be nice for us.

thomasbaldwin:
@anilanar @clucaseh - Hey all, that sounds good I'll work on this :)

danieliser:
How can we remove the async modules created by this plugin entirely?

Our use case is not an app or a library, its a WP plugin. We want to be able to bundle our CSS from components without having a separate entry point for each CSS chunk.

We do this perfectly with this plugin and a definition of splitchunks.

But if each splitChunk has the enforce option set to true, which is required to split the chunks into separate files, then our JS bundle never initiates.

On comparison it now has a `webpackJsonpCallback` block at the top of the bundle and is referencing `deferredModules.push(["./src/index.js","block-editor-styles","block-styles","vendors~block-editor"])`.

The first is our main chunk which is output as `block-editor.js` so shouldn't be async loaded, and the other 3 are caused by mini-css-extract.

If this option allows disabling that output to our JS bundle, then I disagree with the premise it should be relabeled for only preventing output to a browser of a link tag. We don't use any of that as we bundle to package as a release and manually enqueue our JS/CSS assets via the WordPress existing dependency mechanisms.

So there is a need for the originally described solution, the slimmed down version doesn't actually correct the underlying issue.

Sure in most cases actually loading your css async is great, but when given the constraint we are working inside another app, that simply isn't sufficient. Paths to the css/scripts can easily change from site to site making it not viable at all.

Here is sample code we are working from: https://github.com/PopupMaker/Block-Playground/blob/develop/webpack.config.js

As is the current code above outputs multiple CSS bundles correctly, but also builds that extra `webpackJsonpCallback` functionality into our final JS build.

evilebottnawi:
Close in favor https://github.com/webpack-contrib/mini-css-extract-plugin/pull/432, feature will be in near future (this week hope)

PR #353, opened 2019-02-21
feat: add option to disable plugin warnings

AlexDubok:
An option that prevents emitting conflicting order warnings

This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

On a large project with a lot of internal dependencies, there may be a lot of conflicting order warnings. 
Sometimes it is OK, because conflicting files may not contain intersecting selectors. 
Disabling warnings in webpack stats config will not turn off these, so it would be great to have an option to disable plugin's warnings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,6 +272,19 @@ module.exports = {
 
 For long term caching use `filename: "[contenthash].css"`. Optionally add `[name]`.
 
+#### Disabling conflicting order warnings
+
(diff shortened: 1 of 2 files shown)


AlexDubok:
Thanks, didn't know there is an option for that. Closing the PR :)

PR #350, opened 2019-02-18
Uglify -> Terser

evenstensberg:

PR #360, opened 2019-03-08
Update default JS minimizer to Terser

hectorgrebbell:
This PR contains a:

- [x ] **metadata update**

### Motivation / Use-Case

Webpack has moved to Terser so it feels like the documentation here should too

https://github.com/webpack/webpack/pull/8036


hectorgrebbell:
Duplicate of #350

Issue #250, opened 2018-08-22
New version 0.4.2 capture lot of warnings


nuintun:
![_20180822095000](https://user-images.githubusercontent.com/1730277/44438489-edcaea80-a5f1-11e8-85a6-d1f1e5120ec2.png)

Lot of conflicting order warning, but this css files not care about their order!

luqimin:
experiencing the same issue

evilebottnawi:
Due https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/index.js#L459

jsg2021:
What's causing this? What is it checking?

sokra:
We are trying to generate a CSS file but your codebase has multiple possible orderings for these modules.

In this case you used `table` before `button` in one location and `button` before `table` in another location. We need to generate a single CSS file for these and can't decide which one should be first. In this case we put `table` before `button`. But this is chosen without reason. In future build we may decide the other way. But since CSS cares for rule order this can lead to issue when CSS changes without reason.

So not defining a clear order can lead to fragile builds, that's why we display a warning here.

There are two things that you can do:

Sort your imports to create a consistent order. I must admit that the warning could be improved to point to the locations that need to be changed.

Ignore the warning with `stats.warningFilter` when the order of these styles doesn't matter. Maybe the eventually add an option to mark order-independent styles directly.

evilebottnawi:
@wende60 you should solve order problem, not downgrade

wende60:
@evilebottnawi unfortunally i do not even understand what i have to order here... the loaders? Maybe you can give me a hint...

evilebottnawi:
@wende60 looks above and read post https://github.com/webpack-contrib/mini-css-extract-plugin/issues/250#issuecomment-415345126

timothywisdom:
We used the sort-import plugin for VS Code to insure our entire team is importing modules in the same order.
https://marketplace.visualstudio.com/items?itemName=amatiasq.sort-imports

Unfortunately, even after insuring the order is the same, we still see these warnings from the mini-css-extract-plugin

michael-ciniawsky:
@timothywisdom Sorting imports/modules **per file** (within) !== Import/Module Order in the Import/Module Tree/Graph (in relation to **other files**)

jsg2021:
[...]

Issue #362, opened 2019-03-14
ignoreOrder option


skrobek:
Hello!

I'm using the newest package version `0.5.0` which I use to replace deprecated `extract-text-webpack-plugin` with webpack 4.

With the old plugin, we had `ignoreOrder: true` inside our config (needed for CSS) to have our styles loaded in the correct way. After we upgraded to webpack4 and mini-css-extract-plugin I can't find the related option and our CSS modules are not loaded in the correct way. Is there a similar option that might fix our problem ?

evilebottnawi:
No similar option, you need fix order, please provide minimum reproducible test repo

hedgepigdaniel:
As has been pointed out in https://github.com/webpack-contrib/mini-css-extract-plugin/issues/250:
- Sorting all imports (not just in files but in tree traversal import order is extremely difficult to do in a large project
- Consistent use of scoping or naming conventions in CSS makes the order of CSS import irrelevant, and this is a much more manageable strategy for a large project as @jsg2021 said in https://github.com/webpack-contrib/mini-css-extract-plugin/issues/250#issuecomment-416710488
- Using `stats.warningFilter` as recommended by @sokra in https://github.com/webpack-contrib/mini-css-extract-plugin/issues/250#issuecomment-415345126 or [webpack-filter-warnings-plugin](https://github.com/mattlewis92/webpack-filter-warnings-plugin) is not a good solution. It would be much better to avoid generating the warnings at their source (optionally) rather than to partially filter them later with regexes. I use [webpack-error-on-warnings-plugin](https://github.com/hedgepigdaniel/webpack-error-on-warnings-plugin) to enforce that there are no warnings in a build. This is incompatible with filtering warnings, because it must happen in the `before-emit` hook so that it prevents emission of assets when there are warnings. The warnings filtering happens in the `done` hook or in the `toJson` method, which is too late.

evilebottnawi:
Feel free to send a PR `ignoreOrder: true` to avoid problem and warnings

evilebottnawi:
Also we need update docs about usage `warningFilters` for avoid error in output

PR #366, opened 2019-03-26
feat: add orderWarnings flag

hedgepigdaniel:
The flag defaults to true, which retains the existing behaviour of
warnings when there is conflicting import order between multiple CSS
files. When set to false, these warnings are not generated.

This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

As has been pointed out in https://github.com/webpack-contrib/mini-css-extract-plugin/issues/250:
- Sorting all imports (not just in files but in tree traversal import order is difficult to do in a large project, and impossible if there are async chunks with CSS.
- Consistent use of scoping or naming conventions in CSS makes the order of CSS import irrelevant, and this is a much more manageable strategy for a large project as @jsg2021 said in https://github.com/webpack-contrib/mini-css-extract-plugin/issues/250#issuecomment-416710488
- Using `stats.warningFilter` as recommended by @sokra in https://github.com/webpack-contrib/mini-css-extract-plugin/issues/250#issuecomment-415345126 or [webpack-filter-warnings-plugin](https://github.com/mattlewis92/webpack-filter-warnings-plugin) does not work consistently. It's much better to avoid generating the warnings at their source (optionally) rather than to partially filter them later with regexes. For example, I use [webpack-error-on-warnings-plugin](https://github.com/hedgepigdaniel/webpack-error-on-warnings-plugin) to enforce that there are no warnings in a build. This is incompatible with filtering warnings, because it the error must happen in the `before-emit` hook so that it prevents emission of assets when there are warnings. The warnings filtering happens in the `done` hook or in the `toJson` method, which is too late.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,8 @@ module.exports = {
       // Options similar to the same options in webpackOptions.output
       // both options are optional
       filename: "[name].css",
-      chunkFilename: "[id].css"
+      chunkFilename: "[id].css",
+      orderWarning: true // Disable to remove warnings about conflicting order
     })
   ],
   module: {
--- a/src/index.js
+++ b/src/index.js
@@ -113,6 +113,7 @@ class MiniCssExtractPlugin {
     this.options = Object.assign(
       {
         filename: '[name].css',
+        orderWarning: true,
       },
       options
     );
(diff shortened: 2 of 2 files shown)


evilebottnawi:
`stats.warningFilter` should work please provide minimum reproducible test repo where it is doesn't work as expected

hedgepigdaniel:
> For example, I use webpack-error-on-warnings-plugin to enforce that there are no warnings in a build. This is incompatible with filtering warnings, because it the error must happen in the before-emit hook so that it prevents emission of assets when there are warnings. The warnings filtering happens in the done hook or in the toJson method, which is too late.

I can make a minimum repo of this problem but I don't think it will help. `stats.filterWarnings` only filters the warnings that are printed - it doesn't filter the warnings in the compilation object, which is what my plugin uses. `stats.warningFilter` works as intended, it just doesn't solve my problem. It should be clear why if you look at the[ code](https://github.com/hedgepigdaniel/webpack-error-on-warnings-plugin/blob/master/index.js) for webpack-error-on-warnings-plugin. Why put the warnings into the compilation in the first place if you are going to filter them out later?

I think its good for a project to disallow warnings (e.g. in production, the master branch, etc). That way warnings are useful in development. If there's heaps of warnings all the time, nobody reads them or notices new ones and they are not useful. Sometimes that means making a compromise - some warnings are too hard to fix and the fix doesn't provide enough benefit. It's better to hide them than to have heaps of warnings that nobody reads.

All this change does is provide an (optional, non-default) way of turning the warning off in a simpler and cleaner way than filtering strings with regexes.

evilebottnawi:
No it is bad DX. You may miss a problem when you actually have to solve it.

hedgepigdaniel:
Then why are you recommending to use `stats.warningFilter`? That would still make people not see the warning right?

https://github.com/webpack-contrib/mini-css-extract-plugin/issues/250#issuecomment-416710488:
> For example, in my projects each component's styles are self contained so order is only important within each component. (each component has a very unique class name root and all selectors are scoped to them)

In projects that do that, it doesn't matter what order CSS is imported, and the warning is wasting their time.

Having the warning on by default is fine, I can see that it is a potential problem people should be aware of. But not letting them turn it off in a clean way seems counter productive.

pnarielwala-tc:
https://github.com/webpack-contrib/mini-css-extract-plugin/issues/362#issuecomment-480254913

Should this PR be updated with `ignoreOption: true`? Would really love this, thanks!

evilebottnawi:
PR welcome

hedgepigdaniel:
Closing in favour of https://github.com/webpack-contrib/mini-css-extract-plugin/pull/422

Issue #367, opened 2019-03-27
URLs in CSS should be relative to the CSS file, but can't achieve with publicPath


karlvr:
This appears to be an ongoing and reasonably common issue. See #44.

The issue is that webpack outputs URLs relative to the context (assumed to be where the HTML is), which works fine for JS but not CSS, as urls in CSS files are resolved relative to the CSS file, while URLs in JS files are resolved relative to the HTML.

Using `publicPath` is a blunt instrument that either requires you to use an absolute path, or to use a relative path and to always output your CSS files at the same depth (e.g. set `publicPath` to `../` and always output your css files into `css/`).

I am preparing a simple patch for `mini-css-extract-plugin` to add an additional option which I think I'll call `urlsRelativeToSource`. It dynamically adjusts the `publicPath` that it sends to downstream loaders to the relative path from the source CSS file to the root context...

So if the source file is `./forum/css/main.css` then `publicPath` is set to `../../` when passed down to `url-loader` or `file-loader` or whatever is next, which means a url in your CSS that was `../img/header.png`, becomes `../../forum/img/header.png`. When loaded from your CSS _assuming it's output to `./forum/css/main.css`, that relative URL (while containing a redundant `..` and `forum`) works.

The assumption that is required here is that you'll _output_ your CSS files at the same _depth_ in the directory structure as the source file is. Hence the option is `urlsRelativeToSource`. This is because I don't think a loader knows where the output file is going, otherwise we could just make the URLs relative to the output.

Is anyone interested? This fixes the issue for me... will this be useful for others and will it be acceptable as a patch? I don't want to base our workflow on a fork!!

evilebottnawi:
Why they should be relative? It is out of scope this plugin you can use minificator for this.

evilebottnawi:
Also it can potentially break code, because you can use same css in difference places (so resolution `url` was wrong). We already supported this option for `file-loader` in past, but remove this because it is break code in many cases. Here same problem. I strongly recommend don't do this. It is bad practice.

Anyway can you provide example what you actually have and what you expected? Maybe i don't understand you correctly.

karlvr:
I've pushed a branch that implements this and it's working nicely for me so far https://github.com/karlvr/mini-css-extract-plugin/tree/publicPathRelativeToSource

karlvr:
[...]

PR #368, opened 2019-03-27
Add publicPathRelativeToSource option

karlvr:
This PR contains a:

- [ ] **bugfix**
- [X] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Addresses #367 by adding a `publicPathRelativeToSource` option to dynamically adjust the `publicPath` passed down to import loaders so that paths in the output CSS will be relative to the source CSS file rather than the context.

The browser treats URLs in CSS files relative to the CSS file, so having URLs be relative to the CSS file is natural in CSS. A common solution is to set `publicPath` to an absolute path, but that isn't always possible. Another common solution is to set `publicPath` to `../`, which assumes that your CSS file is output into `./css/` (or some other folder, but just one folder) and fails if you output a CSS file more than one folder deep.

(URLs in the CSS that comes _into_ `mini-css-extract-plugin` are made relative to the context by `css-loader`, which produces JavaScript where URLs should be relative to the context)

Note that this solution only works if the output CSS is at the same depth relative to the context as the source file.

### Breaking Changes

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,46 @@ module.exports = {
 }
 ```
 
+#### URLs relative to the CSS file
+
+URLs in CSS files are resolved by the browser relative to the CSS file itself. The `publicPath` option (above) is used
+to specify where the root of the context is, relative to the CSS file for URLs in the CSS to work. However, having just
+one `publicPath` means that all of your CSS files need to be at the same depth (number of directories deep).
+
+Use the `publicPathRelativeToSource` option to dynamically change the `publicPath` used by the `mini-css-extract-plugin` to
+be relative to your source CSS file. NB: in order for the output to be correct, your output CSS file must be at the same 
+relative depth as your source CSS file.
+
+```js
+const MiniCssExtractPlugin = require("mini-css-extract-plugin");
+module.exports = {
+  plugins: [
+    new MiniCssExtractPlugin({
+      // Options similar to the same options in webpackOptions.output
+      // both options are optional
+      filename: "[name].css",
+      chunkFilename: "[id].css"
+    })
+  ],
+  module: {
+    rules: [
+      {
+        test: /\.css$/,
+        use: [
+          {
+            loader: MiniCssExtractPlugin.loader,
+            options: {
+              publicPathRelativeToSource: true,
+            }
+          },
+          "css-loader"
+        ]
+      }
+    ]
+  }
+}
+```
+
 #### Advanced configuration example
 
 This plugin should be used only on `production` builds without `style-loader` in the loaders chain, especially if you want to have HMR in `development`.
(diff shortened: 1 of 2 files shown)


karlvr:
@evilebottnawi I shall put together an example!

I can't use `__webpack_public_path__` as I output multiple CSS files, nested at different depths, e.g. `css/main.css` and `css/admin/main.css` so unless I use an absolute path in `publicPath` or `__webpack_public_path__` there isn't one-size to fit all.

Example to come...

karlvr:
[mcep-pprts-example.zip](https://github.com/webpack-contrib/mini-css-extract-plugin/files/3016048/mcep-pprts-example.zip)

Here is a zip of a very basic project that demonstrates the issue. `publicPath` is set to `../` as I don't want to / can't use an absolute `publicPath`.

The project is setup to use my fork with the code in this PR, so you can turn on/off `publicPathRelativeToSource` in the `webpack.config.js`.

```
npm install
npx webpack
```

Then look at `build/css/main.css`, this has a reference to the image that will work: `../img/background.png`.

Then look at `build/css/admin/admin.css`, this has a reference that won't work, also `../img/background.png` but it needs to be `../../img/background.png`.

This PR would made that relative URL `../../img/background.png`.

Edit `webpack.config.js` and uncomment the line with `publicPathRelativeToSource` and run webpack again and observe the change. Both CSS files now have valid relative paths.

evilebottnawi:
Please search issue in file loader about this and read why it was removed and can break your code in many cases, it is really unsafe, I try to search other approach to solve your problem, but maybe it is won't fix, other bundlers also don't do this, even you do fork it is solution only for you case in your project and can lead to problem in future

karlvr:
@evilebottnawi thank you, I looked at file-loader history and I found mentions of breaking but I didn’t hit upon the concrete example. 

File-loader is indeed the wrong place for this approach, as it outputs many different file types while this module only outputs one. 

In my opinion, every solution that ends up with setting a relative `publicPath` is the wrong approach, as it requires all CSS files at the same depth. So I hope we can work out an appropriate solution and not be left with that!

Are you able to point me to the example where this approach fails (aside from the failure case I identified regarding output being at a different depth than source)? That would really help me understand the problem that I’m not currently seeing clearly!

evilebottnawi:
I think support `publicPath` as function should solve you issue and it is more generic solution, what do you think?

karlvr:
@evilebottnawi it will definitely solve my issue, and I will make a PR for that as it will be nice to have this solvable. I do think that given the interest in relative `publicPath` that actually having a solution as easy as a `boolean` configuration option is useful! I just wish the loader could know where its output was going!

evilebottnawi:
@karlvr feel free to update/create new PR :+1:

evilebottnawi:
Can we close this PR?

karlvr:
All good. I've closed it. I'm going to continue to tinker and see if I can get access to the output file path nicely and do the relative paths comprehensively; for myself and I'll try to share again :-)

Issue #370, opened 2019-03-29
Support insertInto option similar to style-loader?


brophdawg11:
Would this package be open to adding support for an [`insertInto`](https://github.com/webpack-contrib/style-loader#insertinto) option similar to what `style-loader` supports, which allows clients to specify a selector for the DOM node into which they want `link` tags inserted inside of?  By default, it would be the `head`.

We have a scenario where some orchestrated content being injected into an existing app inside of `<body>` and that content contains it's main `app.css` file.  However, any async-loaded CSS files break the expected cascade because they go into `<head>`.  So basically, we end up with a DOM such as:

```
<html> 
  <head>
    ...
    <!-- Injected asyncronously -->
    <link rel="stylesheet" href="async.css" />
  </head>
  <body>  
    ...
    <!-- Orchestrated content rendered server side on page load -->
    <div class="orchestrated-content">
        <link rel="stylesheet" href="app.css" />
        ...
    </div>
    ...
  </body>
</head>
```

When the normal order should be `app.css` -> `async.css`.

I'm planning to get a PR together, but the solution is fairly straightforward it seems.  I've modified `mainTemplate.hooks.requireEnsure.tap` locally to essentially generate JS code that does the following, and it seems to work without issue:

```js
// Assuming no config option passed
var parent = document.querySelector("head");
if (parent) { parent.appendChild(linkTag); }
else { reject("Invalid parent element"); }

// If the user passes config option: { insertInto: ".style-container" }
var parent = document.querySelector(".style-container");
if (parent) { parent.appendChild(linkTag); }
else { reject("Invalid parent element"); }
```

evilebottnawi:
Yep, we can implement this :+1: so it is feature

ScriptedAlchemy:
@brophdawg11 would the style need to be injected in the right order as well? so `async.css` go under `app.css`

jens-duttke:
A functionality like the `insertAt` of style-loader would make more sense, since you can specify the position even more exact.

I have the use-case that I have a large header (different device- and SEO-specific meta tags which are less important than the CSS references).
So I want to inject the CSS-Links directly after the title-tag.

Using
```
    insertAt: {
        after: 'title'
    }
```
that would be easily possible.

To achive the usecase of @brophdawg11, what about:
```
    insertAt: {
        appendTo: '.style-container'
    }
```
?

I think  `appendTo`, `before` and `after`, would cover all use-cases.

[...]

PR #371, opened 2019-04-01
feat(ordering): add support for insertInto option

brophdawg11:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

The issue being addressed is describe in #370.  This adds a new `insertInto` option that is optional and backwards compatible.  Specifying a CSS selector allows the user to specify a DOM node into which the async-loaded `<link>` tags should be inserted.

### Breaking Changes

n/a, changes are backwards compatible and will continue to insert `<link>` tags into the `<head>` if no `insertInto` option is specified

### Additional Info

This is a different approach that solves (I think) the same problem as #331, in a more customizable and backwards-compatible way.

This is also seemingly related to #328.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,6 +356,23 @@ const miniCssExtractPlugin = new MiniCssExtractPlugin({
 
 For long term caching use `filename: "[contenthash].css"`. Optionally add `[name]`.
 
+#### Determining Insertion Point for Async CSS Chunks
+
+If you are extracting CSS into separate files and using Webpack Dynamic Imports, it is possible you will need to be able to configure the insertion point of your async CSS chunks in the document to preserve ordering for the CSS cascade.  This is possible using the `insertInto` option, which takes a function that will be called to determine the parent node into which to insert the `<link>`` tag for the chunk CSs file
+
+```javascript
+module.exports = {
+  ...
+  plugins: [
+    new MiniCssExtractPlugin({
+      filename: "[name].css",
+      insertInto: href => document.querySeletor('.async-styles-container'),
+    })
+  ],
+  ...
+}
+```
+
 ### Remove Order Warnings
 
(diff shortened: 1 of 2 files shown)


brophdawg11:
@evilebottnawi So just do something like `.toString()` on the incoming function to pass it through to the bundled output?  Is that safe for all function types?  I checked quickly, and this does evaluate properly in console:

```
(function foo() { console.log('hello world!'); }).toString()
// "function foo() { console.log('hello world!'); }"
```

If so, I can make that update

evilebottnawi:
@brophdawg11 yes, we do same for `style-loader` and no problem, also it is allow toy write custom logic and it is more generic solution, thanks for PR

brophdawg11:
@evilebottnawi Sorry for the delay - I just updated this to take a function instead of a selector based on your comment above

philipp-spiess:
`style-loader` recently made the custom insertion function a bit more generic by not requiring it to return a parent to insert the style into but instead allow the user to handle the DOM attachment on their own: https://github.com/webpack-contrib/style-loader/pull/413/files

I think it would be great if we can have the same behavior for `mini-css-extract-plugin`. What do you think?

giuseppeg:
👋 folks, I put up a similar PR to introduce the `insert` option as per @evilebottnawi's request https://github.com/webpack-contrib/mini-css-extract-plugin/pull/459 The API is the same as style-loader  and a bit more flexible.  @brophdawg11 I hope that's fine with you (I don't mean to diminishing your good work!)

evilebottnawi:
Close in favor https://github.com/webpack-contrib/mini-css-extract-plugin/pull/459

brophdawg11:
Yep - thanks @giuseppeg!  I agree this is more flexible and better aligns with `style-loader`'s API.  I was able to switch back to the normal package (from my fork) and use `insert` without issue 👍

Issue #377, opened 2019-04-13
[hmr] reloadStyle will trigger errors


hsxfjames:
* Operating System: Mac OS 10.14.4
* Node Version: v11.6.0
* NPM Version: 6.9.0
* webpack Version: 4.30.0
* mini-css-extract-plugin Version: 0.6.0

### Expected Behavior

If we have a `link` in document which its `href` attribute is in data url form, `reloadStyle()` function shouldn't select and parse it. For example:
```html
<link rel="shortcut icon" href="data:;base64,=">
```

### Actual Behavior

`reloadStyle()` function will select this `link` element and try to parse its `href`, then throws an error from `getReloadUrl()` function.
```
Uncaught TypeError: Failed to construct 'URL': Invalid URL
```

### Code

no need?

### How Do We Reproduce?

Add the `link` mentioned above to any html template.

ScriptedAlchemy:
Looking into it now, hope to have PR ready soon. This weekend is a little busy

ScriptedAlchemy:
@hsxfjames do you want to give this a try. https://github.com/webpack-contrib/mini-css-extract-plugin/pull/378

evilebottnawi:
Fixed https://github.com/webpack-contrib/mini-css-extract-plugin/pull/378

hsxfjames:
Not works, because `getReloadUrl()` at line#148 goes before `if (!isUrlRequest(url))`.

Consider to judge `href` in raw? Or how about matching `link`s those with `rel=stylesheet` only?

hsxfjames:
Oh, am I late?😂

evilebottnawi:
`rel` can be omitted, what is problem?

evilebottnawi:
@hsxfjames fix doesn't work for you?

hsxfjames:
@evilebottnawi
Not works, it still throws errors when I try to modify css. The error is from `reloadStyle` -> `getReloadUrl` -> `normalizeUrl` -> `new URL()`. 
I am in a limited network, so I'm also hard to paste full error message here. But you can easily try to reproduce this, just to add a simple link in base64 url form into any template you like. Let me know if you need other information~

evilebottnawi:
@hsxfjames thanks
/cc @ScriptedAlchemy

ScriptedAlchemy:
@hsxfjames will work on this tonight, then ask you to test it again

ScriptedAlchemy:
[...]

PR #378, opened 2019-04-13
fix: do not attempt to reload base64 encoded urls

ScriptedAlchemy:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

No not try and reload non-URL based `<link>` tags

example: `<link rel="shortcut icon" href="data:;base64,=">`
### Additional Info
fixs #377

### DIFF
--- a/src/hmr/hotModuleReplacement.js
+++ b/src/hmr/hotModuleReplacement.js
@@ -90,6 +90,10 @@ function updateCss(el, url) {
     url = el.href.split('?')[0];
   }
 
+  if (isDataURL(url)) {
+    return;
+  }
+
   if (el.isLoaded === false) {
     // We seem to be about to replace a css link that hasn't loaded yet.
     // We're probably changing the same file more than once.
@@ -143,6 +147,10 @@ function reloadStyle(src) {
   forEach.call(elements, function(el) {
     var url = getReloadUrl(el.href, src);
 
+    if (isDataURL(url)) {
+      return;
+    }
+
     if (el.visited === true) {
       return;
     }
@@ -168,6 +176,13 @@ function reloadAll() {
   });
 }
 
+function isDataURL(s) {
+  return !!s.match(
+    // eslint-disable-next-line no-useless-escape
+    /^\s*data:([a-z]+\/[a-z0-9-+.]+(;[a-z-]+=[a-z0-9-]+)?)?(;base64)?,([a-z0-9!$&',()*+;=\-._~:@\/?%\s]*)\s*$/i
+  );
+}
+
 module.exports = function(moduleId, options) {
   if (noDocument) {
     console.log('no window.document found, will not HMR CSS');
@@ -179,6 +194,7 @@ module.exports = function(moduleId, options) {
 
   function update() {
     var src = getScriptSrc(options.filename);
+
     var reloaded = reloadStyle(src);
 
     if (options.locals) {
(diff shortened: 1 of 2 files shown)


src/hmr/hotModuleReplacement.js
```diff
@@ -168,6 +176,13 @@ function reloadAll() {
   });
 }
 
+function isDataURL(s) {
+  return !!s.match(
+    // eslint-disable-next-line no-useless-escape
+    /^\s*data:([a-z]+\/[a-z0-9-+.]+(;[a-z-]+=[a-z0-9-]+)?)?(;base64)?,([a-z0-9!$&',()*+;=\-._~:@\/?%\s]*)\s*$/i
```
evilebottnawi: oh, let's simplify this, no need long regexp, they decrease perf.

Example what we use https://github.com/webpack/loader-utils/blob/master/lib/isUrlRequest.js#L9

https://github.com/webpack/loader-utils/blob/master/lib/isUrlRequest.js#L14

Because we can have `#hash`, `file:`, `http:`/`https:` (absolute urls), schema less urls `//something.com` and `chrome-extension:`, i think we should ignore their too

I can accept this without tests, but will be great have tests for this to avoid future regressions

ScriptedAlchemy: So could a modified version like the following be acceptable?
```function isUrlRequest(url) {
  // An URL is not an request if

  if (/^[a-z][a-z0-9+.-]*:/i.test(url)) {
    return false;
  }

  // 2. It's a protocol-relative
  if (/^\/\//.test(url)) {
    return false;
  }

  // 3. It's some kind of url for a template
  if (/^[{}[\]#*;,'§$%&(=?`´^°<>]/.test(url)) {
    return false;
  }

  // 4. It's also not an request if root isn't set and it's a root-relative url
  if (/^\//.test(url)) {
    return false;
  }

  return true;
}
```

evilebottnawi: No need `4` and for `3` only `#` character

ScriptedAlchemy: So this? :) 
```function isUrlRequest(url) {
  // An URL is not an request if

  // 1. It's an absolute url
  if (/^[a-z][a-z0-9+.-]*:/i.test(url)) {
    return false;
  }

  // 2. It's a protocol-relative
  if (/^\/\//.test(url)) {
    return false;
  }

  return true;
}```

**Comment by ScriptedAlchemy** [comment]
For tests, I wanted to make some additional changes too, like setting up a new folder for some specific unit tests and get the manual test directory running. 

Because I have limited time this weekend, I'm not sure I'll be able to set those parts up. Today or tomorrow

File: src/hmr/hotModuleReplacement.js (hunk 1 above)
Comment (evilebottnawi) [reply src/hmr/hotModuleReplacement.js]: and :
```js
// 3. Avoid `#`
if (/^#/.test(url)) {
  return false;
}
```

To avoid `link="#test"` (based on spec it is allowed).

ScriptedAlchemy:
released this via https://github.com/faceyspacey/extract-css-chunks-webpack-plugin as a precaution :)

evilebottnawi:
@ScriptedAlchemy can you remove draft?

PR #380, opened 2019-04-18
Remove empty modules

### LINKED ISSUE #357: [Optimization request] Remove empty "modules"
When using this plugin we might end up with something like this:
<details>
<summary>Click</summary>
/***/ 54:
/***/ (function(module, exports, __webpack_require__) {

// extracted by mini-css-extract-plugin

/***/ }),

/***/ 92:
/***/ (function(module, exports, __webpack_require__) {

// extracted by mini-css-extract-plugin

/***/ }),

/***/ 93:
/***/ (function(module, exports, __webpack_require__) {

// extracted by mini-css-extract-plugin

/***/ }),

/***/ 94:
/***/ (function(module, exports, __webpack_require__) {

// extracted by mini-css-extract-plugin

/***/ }),

/***/ 95:
/***/ (function(module, exports, __webpack_require__) {

// extracted by mini-css-extract-plugin

/***/ }),

/***/ 96:
/***/ (function(module, exports, __webpack_require__) {

// extracted by mini-css-extract-plugin

/***/ }),

/***/ 97:
/***/ (function(module, exports, __webpack_require__) {

// extracted by mini-css-extract-plugin

/***/ }),

/***/ 98:
/***/ (function(module, exports, __webpack_require__) {

// extracted by mini-css-extract-plugin

/***/ }),
</details>

Seems like this could be optimized away - i know that probably minifier + gzip makes this kinda cheap, but parsing costs add up and it would be always good to strive for less code in such cases.

irudoy:
This PR contains a:

- [x] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #357


evilebottnawi:
Please fix CI problems, also accept CLA, thanks

src/index.js
```diff
@@ -16,6 +16,18 @@ const pluginName = 'mini-css-extract-plugin';
...
+const REGEXP_COMMENTS = /\/\*[\s\S]*?\*\/|([^\\:]|^)\/\/.*$/gm;
+
+const moduleHasNoSource = (module) => {
+  // eslint-disable-next-line no-underscore-dangle
+  if (!module._source || !module._source._value) return true;
+
+  return (
+    // strip all comments
+    // eslint-disable-next-line no-underscore-dangle
+    module._source._value.replace(REGEXP_COMMENTS, '').trim().length === 0
```
evilebottnawi: `_*` options is private and can be removed in future or between versions, we need search way how we can do this without using private properties

irudoy:
@evilebottnawi do you suggest updating vulnerable dependencies in this PR?
CLA has already been signed.

src/index.js
irudoy: Fixed

src/index.js
```diff
@@ -16,6 +16,22 @@ const pluginName = 'mini-css-extract-plugin';
...
+  if (typeof module.originalSource !== 'function') return true;
+
+  const source = module.originalSource();
+
+  return Boolean(
+    source &&
+      source
+        .source()
+        // strip all comments
+        .replace(REGEXP_COMMENTS, '')
```
evilebottnawi: It is for `// extracted by mini-css-extract-plugin`?

irudoy: Yep. In order not to remove the modules with valid JS (HMR, locals), to be precise.

evilebottnawi: :+1: I think we don't need regexp here let's check module source on string `// extracted by mini-css-extract-plugin` and all

irudoy: In this case, we should reuse `pluginName` and `// extracted by ${pluginName}` constants between loader and plugin code. Otherwise, we cannot be sure that they will be the same. At the moment, the code is built using only the Babel, and we are not able to require some shared module.

Anyway, I can reluctantly do this, just let me know 😉

evilebottnawi: @irudoy I think we can solve this more simple way - when we replace content on comment we can add property like `module.extracted` and remove only modules what contains this property. Using regexp is bad idea, because other plugin can modify code too and regexp can not works in all cases.

irudoy: Like this?

```javascript
// loader.js
// ...

let resultSource = `// extracted by ${pluginName}`;
let extracted = true;

if (locals) {
    resultSource += `\nmodule.exports = ${JSON.stringify(locals)};`;
    extracted = false;
}

if (options.hmr) {
    resultSource = hotLoader(resultSource, {
        context: this.context,
        options,
        locals,
    });
    extracted = false;
}

this._module.buildMeta = { ...this._module.buildMeta, extracted };

return callback(null, resultSource);
```

evilebottnawi: Yep :+1:

irudoy: Done

irudoy:
Friendly ping

evilebottnawi:
All looks very good, wait @sokra review (we do this in near future), thanks fro waiting

aberezkin:
@sokra, @evilebottnawi, any news about that?

src/index.js
```diff
@@ -432,6 +432,36 @@ class MiniCssExtractPlugin {
           return source;
         }
       );
+
+      compilation.hooks.optimizeChunks.tap(pluginName, (chunks) => {
+        for (const chunk of chunks) {
+          chunk.modulesIterable.forEach((module) => {
```
sokra: An iterable has no forEach method. You should not assume that it's an Set.

Use `for of` instead.

src/index.js
```diff
@@ -432,6 +432,36 @@ class MiniCssExtractPlugin {
...
+            // eslint-disable-next-line no-param-reassign
+            module.dependencies = module.dependencies.filter((dep) => {
+              const hasCssDep =
+                dep.module &&
+                dep.module.dependencies.find(
```
sokra: The whole loop is very expensive. `chunks * modules * dependencies * dependencies`. I don't think this is acceptable for larger builds.

src/index.js
```diff
@@ -432,6 +432,36 @@ class MiniCssExtractPlugin {
...
+                  (d) => d.module && d.module.type === MODULE_TYPE
+                );
+
+              if (hasCssDep) {
+                dep.disconnect();
```
sokra: I think this works only for ESM dependencies. `require()` probably break because of this.

Could you add a CommonJS test cases?

src/index.js
```diff
@@ -432,6 +432,36 @@ class MiniCssExtractPlugin {
...
+                return false;
+              }
+
+              return true;
+            });
+
+            if (
+              module.type === MODULE_TYPE &&
+              module.issuer &&
```
sokra: A module can be referenced by multiple modules, but there is only one issuer.

Avoid using issuer.

src/index.js
ScriptedAlchemy: Definitely not performant. mini-css already does get expensive on large sites with many css files. It would be quite noticeable. 

What if we used something other than pure JS which has pretty bad performance on loops compared to lodash or transducers.js. 

https://jlongster.com/Transducers.js-Round-2-with-Benchmarks

@sokra I have been considering trying to improve loop performance. Any thoughts on an implementation like one mentioned in the article? Benchmarks are impressive with substantial performance over native js functions

irudoy: Made a stress test for this case. Ran three times with these settings: size 100, depth 80. The results are quite interesting:

`npm run test:stress`

```
Mode: development
Modules: 24002
Build time (without loop):  94944ms, 99758ms, 97380ms
Build time (with loop):     92237ms, 98391ms, 94587ms
Assets (without loop):      development.js 15.3 MiB
Assets (with loop):         development.js 6.47 MiB

Mode: production
Modules: 16001
Build time (without loop):  96744ms, 102081ms, 99210ms
Build time (with loop):     92744ms, 98898ms, 94731ms
Assets (without loop):      production.js 203 KiB
Assets (with loop):         production.js 951 bytes (hi!)
```

irudoy: @sokra @ScriptedAlchemy @evilebottnawi please take a look

irudoy:
Any news?

test/cases/simple-require/expected/main.js
```diff
@@ -0,0 +1,95 @@
...
+/******/
+/******/ 	// Load entry module and return exports
+/******/ 	return __webpack_require__(__webpack_require__.s = 0);
+/******/ })
+/************************************************************************/
+/******/ ([
+/* 0 */
+/***/ (function(module, exports, __webpack_require__) {
+
+__webpack_require__('./style.css');
```
sokra: Looks like this will crash at runtime

src/index.js
rjgotten: What about internally using some `Set` or `Map` to memoize previous results and eliminate duplicate evaluation. Would that not help with modules duplicated over multiple chunks and in particular with reducing the tracing for css dependencies?

Somewhat like e.g.

```js
compilation.hooks.optimizeChunks.tap(pluginName, (chunks) => {

  const modules = new Set();
  
  for (const chunk of chunks) {
    for (const module of chunk.modulesIterable) {
      modules.add(module);
      
      if (module.type !== MODULE_TYPE || !module.reasons) break;
      
      for (const reason of module.reasons) {
        if (!reason.module || !reason.module.buildMeta.extracted) break;
        chunk.removeModule(reason.module);
      }
    }
  }
  
  const cssDeps = new Map();
  
  modules.forEach((module) => {
    module.dependencies = module.dependencies.filter((dep) => {
      let hadDep = cssDeps.get(dep);
      if (hadDep == null) {
        cssDeps.set(dep, hadDep = checkForCssDep(dep));
        if (hadDep) { dep.disconnect(); }
      }
          
      return !hadDep;
    });
  });
  
  function checkForCssDep( dep ) {
    return dep.module && dep.module.dependencies.some(
      (d) => d.module && d.module.type === MODULE_TYPE
    );
  }
});
```

src/index.js
irudoy: @sokra this is definitely will break CJS dependencies. Do you have any thoughts on how to avoid that?

src/index.js
irudoy: @sokra this implementation is much more complicated than the original. But performance seems like the same.

The tricky part is that we need to disconnect HarmonyImportSideEffectDependency, but we have only CssDependency.

Should we use this or previous version?

```javascript
class SetMap extends Map {
  set(key, value) {
    return super.set(key, this.has(key) ? this.get(key).add(value) : new Set([value]));
  }
}

// ...

const toRemoveMap = new SetMap();

for (const chunk of chunks) {
  for (const module of chunk.modulesIterable) {
    if (module.type === MODULE_TYPE) {
      for (const cssModuleReason of module.reasons) {
        for (const reason of cssModuleReason.module.reasons) { // 💩
          toRemoveMap.set(reason.module || cssModuleReason.module, reason.dependency);
        }
        if (cssModuleReason.module && cssModuleReason.module.buildMeta.extracted) {
          chunk.removeModule(cssModuleReason.module);
        }
      }
    }
  }
}

for (const [module, set] of toRemoveMap) {
  module.dependencies = module.dependencies.filter(d => {
    if (set.has(d)) {
      d.disconnect();
      return false;
    }
    return true;
  });
}
```

irudoy:
@sokra @evilebottnawi is it ok now?

irudoy:
Any news?

src/index.js
```diff
@@ -21,6 +21,15 @@ const REGEXP_NAME = /\[name\]/i;
 const REGEXP_PLACEHOLDERS = /\[(name|id|chunkhash)\]/g;
 const DEFAULT_FILENAME = '[name].css';
 
+class SetMap extends Map {
+  set(key, value) {
+    return super.set(
```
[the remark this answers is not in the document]
sokra: I would also rename the method to `add` instead of `set`, that makes more sense to me.

src/index.js
```diff
@@ -390,6 +399,47 @@ class MiniCssExtractPlugin {
...
+                      reason.dependency
+                    );
+                }
+
+                if (
+                  !isCjs &&
+                  cssModuleReason.module &&
+                  cssModuleReason.module.buildMeta.extracted
+                ) {
+                  chunk.removeModule(cssModuleReason.module);
```
sokra: Should be unneeded when using `finishModules` hook.

evilebottnawi:
/cc @irudoy you do great job, big thanks, ping me when you need review or help from me or sokra

We do release after merge this

igoradamenko:
Hey guys!

What is the status of the PR? Is it going to be merged? I see that @irudoy has already pushed some changes. Are they enough to merge or not? May I help to speed up the development here in any way?

irudoy:
@igoradamenko as @sokra noted we should implement it using `finishModules` hook. Unfortunately, I haven't much time at the moment to do this. I will try to finish it soon.

therealshark:
@irudoy Did you have any time to work on this? The Project i'm working on currently has ~640 empty modules because of this behaviour, so this pr would help us quite a bit :)

kherock:
For those following this issue that still can't use webpack 5 for one reason or another -

kherock:
For those following this issue that still can't use webpack 5 for one reason or another - I've created a gist that contains the most complete implementation of a plugin for removing empty JS bundles that I'm aware of:  
[Remove empty CSS modules workaround](https://gist.github.com/kherock/69ecd01619c72b3b94806c46342ba8d8)

The implementation is mostly inspired by the work in this PR plus some of the suggested changes. Some of the issues I ran into with other workarounds posted is that they break with split chunks or have other quirks when optimization is enabled.

irudoy:
@kherock thanks for the contribution!

@sokra @evilebottnawi 
As of webpack-5-beta.13, the problem still exists. Dead code with empty modules remains in the production bundle. Maybe it is possible to remove it at the webpack level? If there is no code in the module (excluding comments), remove the module automatically.

If this option is not suitable, maybe we have a new and more proper way to remove unnecessary modules in webpack-5? Or maybe we need the new API as of webpack-5, which will allow us to easily remove unneeded modules?

The current implementation looks fragile and brings an overhead that seems to be avoided by doing so in the webpack core.

ScriptedAlchemy:
if someone has something stable, ill be willing to accept a PR to extract-css-chunks-webpack-plugin. Its got a decent amount of users and I regularly try to PR changes from extract-css back to mini-css. Serves as a good testing ground as well. So if anyone wants something installable on NPM, open a PR there and as long as its stable ill release it till this PR gets merged.

irudoy:
Seems like it was fixed by https://github.com/webpack-contrib/mini-css-extract-plugin/pull/546

Can we close this PR and related issue?

evilebottnawi:
Yes, let's add more tests, WIP

irudoy:
@evilebottnawi with another PR, maybe?

evilebottnawi:
@irudoy yes, we can do it in the separate PR

irudoy:
@evilebottnawi done https://github.com/webpack-contrib/mini-css-extract-plugin/pull/566

aleen42:
I have tested with `{esModule: true}`, which is `true` by default, and found that there were still so many empty modules:

```js
(self["webpackJsonp"] = self["webpackJsonp"] || []).push([["X"],{
    // ...

/***/ "./a.less":
/*!***************************************************************************************************************!*\
  !*** ./a.less ***!
  \***************************************************************************************************************/
/***/ (function() {

eval("// extracted by mini-css-extract-plugin\n\n//# sourceURL=webpack://cmxt.wmweb/./a.less?");

/***/ }),

/***/ "./b.less":
/*!***************************************************************************************************!*\
  !*** ./b.less ***!
  \***************************************************************************************************/
/***/ (function() {

eval("// extracted by mini-css-extract-plugin\n\n//# sourceURL=webpack://cmxt.wmweb/./b.less?");

/***/ }),

    // ...
}]);
```

alexander-akait:
Because you import them in JS, we can't fully remove them, otherwise if will get wrong behavior

aleen42:
> Because you import them in JS, we can't fully remove them, otherwise if will get wrong behavior

The hardest thing to remove is to eliminate all import points in JavaScript, but it should be a feature to do show.

PR #381, opened 2019-04-19
feat(options): add moduleFilename option

kevinchappell:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

The `filename` option currently takes a string the enables devs to customize the filename of the generated css file. This works well and the customized filename is created in the same output directory as the js file it is extracted from, however there are cases when the generated files may need to be in different directories. This change enables modifying the `filenameTemplate` using chunk data.

#### Use Case:
webpack.config.js - `entry`
```javascript
  entry: {
    'dist/main': 'src/index.js',
    'demo/assets/js/demo': 'src/demo/index.js',
  }
```
The above config will create `dist/main.js` and `demo/assets/js/demo.js`.

#### Use:
webpack.config.js - `plugins`
```javascript
  plugins: [
    new MiniCssExtractPlugin({
      moduleFilename: ({ name }) => `${name.replace('/js/', '/css/')}.min.css`,
    })
  ]
```
We can change the output directory, chunkhash length and more by using a function that gets called during `compilation.mainTemplate.hooks.renderManifest`.

### Breaking Changes

None


src/index.js
```diff
@@ -197,7 +195,8 @@ class MiniCssExtractPlugin {
                   renderedModules,
                   compilation.runtimeTemplate.requestShortener
                 ),
-              filenameTemplate: this.options.filename,
+              filenameTemplate:
+                this.options.filenameTemplate || this.options.filename,
```
evilebottnawi: No need new function here you should use `filename: () => { /* logic for nameing */ }`

src/index.js
```diff
@@ -16,6 +16,7 @@ const pluginName = 'mini-css-extract-plugin';
 const REGEXP_CHUNKHASH = /\[chunkhash(?::(\d+))?\]/i;
 const REGEXP_CONTENTHASH = /\[contenthash(?::(\d+))?\]/i;
 const REGEXP_NAME = /\[name\]/i;
+const REGEXP_PLACEHOLDERS = /\[(name|id|chunkhash)\]/g;
```
evilebottnawi: Should be solved in separate PR

src/index.js
kevinchappell: good idea. Should we also only pass `{chunk}`?

src/index.js
kevinchappell: removed

src/index.js
kevinchappell: `filenameTemplate` now called from an anonymous function

evilebottnawi:
Can you rebase, thanks for PR, good work!

kevinchappell:
> Can you rebase, thanks for PR, good work!

done

src/index.js
```diff
@@ -122,7 +123,8 @@ class MiniCssExtractPlugin {
   constructor(options) {
     this.options = Object.assign(
       {
-        filename: '[name].css',
+        filename: DEFAULT_FILENAME,
+        filenameTemplate: () => options.filename || DEFAULT_FILENAME,
```
evilebottnawi: Let's rename `filenameTemplate` to `moduleFilename` (example for wasm https://github.com/webpack/webpack/blob/a4e5f63d9d370b191bb9586570bc22bf2947e390/lib/wasm/WebAssemblyModulesPlugin.js#L65)

kevinchappell: good idea. done

ghostflow:
when will this pr be merged?
need this feature~

ifeltsweet:
This doesn't work on chunks that are split out with SplitChunksPlugin – only works for entries.

PR #384, opened 2019-04-25
fix(loader): fix publicPath regression

rmja:
... when the publicPath is set to an empty string. The publicPath should resolve to an empty string in this case and _not_ `/`.

This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

A regression caused by https://github.com/webpack-contrib/mini-css-extract-plugin/pull/373 makes it impossible to set `publicPath` to an empty string. This fix brings this feature back.


evilebottnawi:
Ignore CI problem , just add tests and we can release new version

evilebottnawi:
Also please change email on commit and do force push, need accept CLA, thanks

rmja:
@evilebottnawi year, sorry about that. I think it is fixed now.

PR #399, opened 2019-05-23
fix: downgrading normalize-url

### LINKED ISSUE #385: normalize-url not working in IE11
This module depends on `normalize-url`, but it doesn't work for IE11 because it's not transpiled to ES5.  
For example https://github.com/sindresorhus/normalize-url/blob/master/index.js#L5 this line contains an arrow function.
I can see that there is a similar package but for es5 https://www.npmjs.com/package/normalize-url-es5.

ScriptedAlchemy:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Downgrading normalize-url to close issues opened around ES5 support required in IE

### Additional Info
Internet Explorer is awesome...

fix #385

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1777,7 +1777,7 @@
     },
     "array-equal": {
       "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
+      "resolved": "http://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
       "integrity": "sha1-jCpe8kcv2ep0KwTHenUJO6J1fJM=",
       "dev": true
     },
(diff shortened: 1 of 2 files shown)


evilebottnawi:
Looks good, in future we should implement build stage for transpile package, but we can avoid this here

evilebottnawi:
Also don't forget move https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/hmr/hotModuleReplacement.js#L145 before `getReloadUrl`, because in some cases `el.href` can be undefined or invalid

evilebottnawi:
You can do this in other PR

ScriptedAlchemy:
Appreciate it. Been out of town this weekend. Will catch up as I do see there has been some activity.

evilebottnawi:
Yes, I written test for home too

Issue #413, opened 2019-06-21
Ability to add attributes to link tag and intercept onload event


Rulexec:
### Feature Proposal

Add ability to add custom code to `requireEnsure` hook code to be able do something custom with the `linkTag`.

For me there is only two cases:

- Add custom code right after `var linkTag = document.createElement("link");` (to be able add some attributes)
- Replace `onload` function to do some work before `resolve()` call

### Feature Use Case

I'm using `css-vars-ponyfill` to use CSS Variables in IE11. And I want to call it after `<link>` tag was added by `mini-css-extract-plugin`, but before chunk will be loaded. Also, I need to add custom attribute to `<link>` tag to handle only own CSS.

Currently I'm solving it in a very dirty way:

```
compilation.mainTemplate.hooks.requireEnsure.tap({
	name: 'SkinsPlugin',
	stage: 9999999,
}, source => {
	const miniCssPluginPresent = source.indexOf('if(installedCssChunks[chunkId]) promises.push(installedCssChunks[chunkId]);') >= 0;
	if (!miniCssPluginPresent) {
		throw new Error('SkinsPlugin:noMiniCssExtractPlugin');
	}

	let replaced = false;

	const regexp =/^(\s*)linkTag.onload = resolve;$/m;

	source = source.replace(regexp, (full, indent) => {
		replaced = true;

		let code = ``;

		code += indent + `linkTag.setAttribute('data-include-css-ponyfill', '');`;

		code += indent + `linkTag.onload = function(e){\n`;
		code += indent + `\tif (typeof SomeGlobalObject !== 'undefined' && SomeGlobalObject && SomeGlobalObject._someFun) {\n`;
		code += indent + '\t\tSomeGlobalObject._someFun();\n';
		code += indent + '\t\tresolve(e);\n';
		code += indent + '\t}\n';
		code += indent + '};\n';

		return code;
	});

	if (!replaced) {
		throw new Error('SkinsPlugin:noMiniCssExtractPlugin');
	}

	return source;
});
```

But I think, that better is to collect some other use cases which need interception and implement some options to do it.

evilebottnawi:
hm, can you create example of this? I want investigate this situation? Looks like a good feature, thanks

Rulexec:
@evilebottnawi sure — https://github.com/Rulexec/mini-css-extract-plugin-css-vars-ponyfill

evilebottnawi:
[...]

PR #418, opened 2019-06-26
feat: add attrs option

fengzilong:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

related issue: https://github.com/webpack-contrib/mini-css-extract-plugin/issues/413

support customizing attributes for link tag

### Breaking Changes

No


fengzilong:
I've signed the CLA, but the bot said `not signed yet`

Is that caused by my second commit? anyone can help?

fengzilong:
Thanks for reply, I'll look into the hooks implementation

dutzi:
Any updates regarding this PR? 

Our use case: we're using css-vars-ponyfill for IE11, we are a widget embedded in a website, we want css-vars-ponyfill to run only on CSS generated by us.

dutzi:
Hey @evilebottnawi, I tried looking at the `test/cases` directory for reference as to how I can write a test for such feature, but I couldn't find anything that resembles this setup. In this test we are required to test for the code generated by [this](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/index.js#L367-L430) block, but what file does it output? What would I put in the `expected` folder

Can you please provide some guidance?

evilebottnawi:
Close in favor https://github.com/webpack-contrib/mini-css-extract-plugin/pull/445

PR #422, opened 2019-07-10
feat: Add ignoreOrder flag

### LINKED ISSUE #362: ignoreOrder option

mjhenkes:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
Picking up the torch where https://github.com/webpack-contrib/mini-css-extract-plugin/pull/366 left off and making the changes suggested by @evilebottnawi 
Also addresses #362
Borrowing @hedgepigdaniel 's description:

As has been pointed out in #250:

* Sorting all imports (not just in files but in tree traversal import order is difficult to do in a large project, and impossible if there are async chunks with CSS.
* Consistent use of scoping or naming conventions in CSS makes the order of CSS import irrelevant, and this is a much more manageable strategy for a large project as @jsg2021 said in #250 (comment)
* Using stats.warningFilter as recommended by @sokra in #250 (comment) or webpack-filter-warnings-plugin does not work consistently. It's much better to avoid generating the warnings at their source (optionally) rather than to partially filter them later with regexes. For example, I use webpack-error-on-warnings-plugin to enforce that there are no warnings in a build. This is incompatible with filtering warnings, because it the error must happen in the before-emit hook so that it prevents emission of assets when there are warnings. The warnings filtering happens in the done hook or in the toJson method, which is too late.

### Breaking Changes
None
### Additional Info
The added tests is to ensure that enabling the flag doesn't make the plugin implode, but it doesn't really programmatically verify the warnings aren't printed.

fixes #362

### DIFF
--- a/test/IgnoreOrder.test.js
+++ b/test/IgnoreOrder.test.js
@@ -1,0 +1,44 @@
+import path from 'path';
+import webpack from 'webpack';
+
+describe('IgnoreOrder', () => {
+  it('should emit errors', (done) => {
+    const casesDirectory = path.resolve(__dirname, 'cases');
+    const directoryForCase = path.resolve(casesDirectory, 'ignoreOrderFalse');
+    // eslint-disable-next-line import/no-dynamic-require, global-require
+    const webpackConfig = require(path.resolve(
+      directoryForCase,
+      'webpack.config.js'
+    ));
+    const compiler = webpack({
+      ...webpackConfig,
+      mode: 'development',
+      context: directoryForCase,
+      cache: false,
+    });
+    compiler.run((err1, stats) => {
+      expect(stats.hasWarnings()).toBeTruthy();
+      done();
+    });
+  });
+
+  it('should not emit errors', (done) => {
+    const casesDirectory = path.resolve(__dirname, 'cases');
+    const directoryForCase = path.resolve(casesDirectory, 'ignoreOrder');
+    // eslint-disable-next-line import/no-dynamic-require, global-require
+    const webpackConfig = require(path.resolve(
+      directoryForCase,
+      'webpack.config.js'
+    ));
+    const compiler = webpack({
+      ...webpackConfig,
+      mode: 'development',
+      context: directoryForCase,
+      cache: false,
+    });
+    compiler.run((err1, stats) => {
+      expect(stats.hasWarnings()).toBeFalsy();
+      done();
+    });
+  });
+});
--- a/test/IgnoreOrder.test.js
+++ b/test/IgnoreOrder.test.js
@@ -2,7 +2,7 @@ import path from 'path';
 import webpack from 'webpack';
 
 describe('IgnoreOrder', () => {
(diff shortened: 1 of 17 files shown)


kubijo:
Can someone from mainteiners merge this please.
The code seems nice, terse and without conflicts, so this should be a quick pass-over

mjhenkes:
@evilebottnawi technically all the other tests are testing ignoreOrder: false, but i've added an explicit test.
I've also updated the package-lock file, it started failing due to npm audit.

mjhenkes:
@evilebottnawi added

test/IgnoreOrder.test.js
```diff
@@ -0,0 +1,44 @@
...
+      'webpack.config.js'
+    ));
+    const compiler = webpack({
+      ...webpackConfig,
+      mode: 'development',
+      context: directoryForCase,
+      cache: false,
+    });
+    compiler.run((err1, stats) => {
+      expect(stats.hasWarnings()).toBeTruthy();
```
evilebottnawi: `toBe(true)`

test/IgnoreOrder.test.js
```diff
@@ -0,0 +1,44 @@
...
+      'webpack.config.js'
+    ));
+    const compiler = webpack({
+      ...webpackConfig,
+      mode: 'development',
+      context: directoryForCase,
+      cache: false,
+    });
+    compiler.run((err1, stats) => {
+      expect(stats.hasWarnings()).toBeFalsy();
```
evilebottnawi: `toBe(false)`

test/IgnoreOrder.test.js
```diff
@@ -0,0 +1,44 @@
+import path from 'path';
```
evilebottnawi: rename file `ignoreOrder-option.test.js`

mjhenkes:
Do we need more approvals?

mjhenkes:
Thanks @evilebottnawi! 🎉

kubijo:
Yeah, thanks a lot guys!
What is the release process now?

evilebottnawi:
Today

test/IgnoreOrder.test.js
JimTheMan: just wondering why toBeTruthy(); is "better". Seems like the exact same shit to me...

JimTheMan: @evilebottnawi

PR #432, opened 2019-07-20
Feature: Disable Extracting Specified CSS Module from Chunks

lsycxyj:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

You may disable extracting css modules programmatically by passing a function.

```javascript
const miniCssExtractPlugin = new MiniCssExtractPlugin({
  disableExtract({ module, isAsync }) {
    // Do whatever you want. Disable by content size for example:
    // return module.content.length < 10 * 1024;
    // Or disable extracting from all async chunks
    return isAsync;
  },
});
```

### Breaking Changes

None

### Additional Info

I found #348 doesn't work at all by now and it's not flexible enough, so I made another pull request based on recent master branch.


lsycxyj:
Well... the lint problems may be a bit difficult to fix...

evilebottnawi:
What is use case? Why you need disable async css?, also i think we should solve this in other case, provide developers function `shouldBeExtracted(options) { if (something) { return false } return true; }`, it is allow solve many cases, not only async, also you can keep small css inside js and many other cases

lsycxyj:
@evilebottnawi See this [comment](https://github.com/webpack-contrib/mini-css-extract-plugin/pull/348#issuecomment-475138619)

evilebottnawi:
@lsycxyj please read my comment above, we can implement general solution instead only for one case

lsycxyj:
@evilebottnawi For general solution, I think a "shouldBeExtracted" function for all modules would be much better. I can decide whether a css module depending on its information: whether it is an async module, content, file size, and more.

evilebottnawi:
@lsycxyj yes, feel free to update PR and implement this option :+1:

lsycxyj:
@evilebottnawi Well... Honestly, I'm not quite familiar with the source code of webpack... Maybe it would be a little difficult for me to implement it for a general purpose... Can you give me some advices to implement it?

lsycxyj:
@evilebottnawi I found this PR would introduce duplicated css dependencies. Is there any other way that I can restore the extracted module without introducing duplicated dependencies?

lsycxyj:
@evilebottnawi I managed it. Now `disableExtract` option has been implemented.

README.md
```diff
@@ -343,23 +343,25 @@ module.exports = {
...
-#### Disable extracting css from async chunks
+#### Disable Extracting Specified CSS from Chunks
 
-You may disable extracting css from async chunks with `disableAsync` option.
+You may disable extracting css from all chunks with `disableExtract` option.
 
 ```javascript
 const miniCssExtractPlugin = new MiniCssExtractPlugin({
-  disableAsync: true,
+  disableExtract: true,
```
evilebottnawi: No need `Boolean` value here, this plugin for extract, and include plugin and disable this is misleading, only `Function` support

evilebottnawi:
Look here https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/loader.js#L101, just skip module for keeping module in js

lsycxyj:
@evilebottnawi Do you mean like this?

README.md
lsycxyj: Boolean type option has been removed.

lsycxyj:
@evilebottnawi Friendly ping

lsycxyj:
Hi, @evilebottnawi, Is there anything else need to be improved? I suppose there can't be less code changes to implement this feature.

evilebottnawi:
In todo, a lot of issues, you can use own branch if need this asap

lsycxyj:
Merge from latest master, fix bug and reopen

lsycxyj:
Known issue: some variables will become "void 0" in some modules after enabling disableExtract in complicated cases, and I don't have any good idea that I can rebuild the modules correctly. Could you do me a favor, please?  @evilebottnawi

README.md
```diff
@@ -499,7 +499,22 @@ module.exports = {
...
 ```
 
-### Module Filename Option
+### Disable Extracting Specified CSS from Chunks
+
+You may disable extracting css modules programmatically by passing a function.
+
+```javascript
+const miniCssExtractPlugin = new MiniCssExtractPlugin({
+  disableExtract({ module, isAsync }) {
```
sokra, line 508: Name it in positive way e.g. should extract

lsycxyj:
> The functionally seem to be based on rebuilding modules. That's really problematic as it leads to inconsistencies and performance problems. Rebuilding modules should be avoided.
> 
> There is already the option to apply mini-css only to some modules via the rule system. I don't think this PR is needed anymore when the `isAsync` feature is removed.

@sokra Thank you for your review. 

Actually, this feature is originally named "disableAsync". The main purpose of the feature is to find out async modules and disable the extraction programactically.

Then, @evilebottnawi gave me advice that to do it with a function like `shouldBeExtracted`, and it became "disableExtract".

I do understand it leads to some problems. However, I can't figure out any better way to implement the feature by now. I'm still looking for an alternative way.

evilebottnawi:
> There is already the option to apply mini-css only to some modules via the rule system. I don't think this PR is needed anymore when the isAsync feature is removed.

Why this is unacceptable?

Issue #424, opened 2019-07-15
Add a configuration flag to enable ES module syntax


joseprio:
# Feature Proposal
At the moment the generated JS module uses the commonJS syntax. It would be useful to have an option that would use the ES module syntax instead.

# Feature Use Case
Old commonJS syntax prevents module concatenation in Webpack. It may help tree shaking in some cases too.

PR #433, opened 2019-07-20
Add support for ES2015 module syntax

joseprio:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update**
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Using the CommonJS module syntax prevents Webpack optimization features from running (tree shaking, module concatenation).

Issue #424

### Breaking Changes

Default is to use the old syntax, so no

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,10 +51,9 @@ module.exports = {
   plugins: [
     new MiniCssExtractPlugin({
       // Options similar to the same options in webpackOptions.output
-      // all options are optional
+      // both options are optional
       filename: '[name].css',
       chunkFilename: '[id].css',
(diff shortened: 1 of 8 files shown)


hiroppy:
Please add tests.

joseprio:
> Please add tests.

@hiroppy I actually wanted to add a test for this, but the current test suite seems broken in master. I checked other PRs, and they weren't including tests in new features. I tested manually and it's doing what it's supposed to do.

joseprio:
@evilebottnawi tests added, please review at your convenience :)

evilebottnawi:
Need do same for `style-loader` too to avoid incompatibility

PR #435, opened 2019-07-23
feat: hmr error recovery

rjgotten:
This PR contains a:

- [ ] **bugfix**
- [X] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

This PR implements error-resilience/-recovery for HMR when dealing with failed CSS child compilations. (E.g. syntax errors in Less or Sass code.)

It enables failed compilations to continue to `hot.accept` future compilations, instead of requiring a full reload. This is a significant improvement to workflow, esp. for larger applications that can't deal well with reload and require manually recreating a certain application state.

### Breaking Changes

None, as far as I am aware. All existing test scenarios in the suite continue to pass.

### Additional Info

The [NetMatch/mini-css-extract-plugin-hmr-testcase](https://github.com/NetMatch/mini-css-extract-plugin-hmr-testcase) repository contains a testcase that illustrates the problem; includes a method to switch to the patched fork and illustrate the improvements; and has a rundown of the how/what/why of changes.

A test update will probably still be required to verify the resilience. But I'm having some problems wrapping my head around how the test suite is set up.
Some assistance there would be appreciated.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -31,9 +31,16 @@ function hotLoader(content, context) {
     locals: !!context.locals,
   })});
       module.hot.dispose(cssReload);
+      module.hot.accept(function(){});
       ${accept}
     }
   `;
+}
+
+function interceptError(callback, interceptor) {
+  return (err, source) => {
+    return callback(null, err ? interceptor(err) : source);
+  };
 }
 
 const exec = (loaderContext, code, filename) => {
@@ -133,15 +140,22 @@ export function pitch(request) {
     });
   });
 
-  const callback = this.async();
+  const callback = !options.hmr
+    ? this.async()
+    : interceptError(this.async(), (err) => {
+        let resultSource = `// extracted by ${pluginName}`;
+        resultSource += hotLoader('', {
+          context: this.context,
+          locals: null,
+          options,
+        });
+        resultSource += `\nthrow new Error(${JSON.stringify(String(err))});`;
+        return resultSource;
+      });
 
   childCompiler.runAsChild((err, entries, compilation) => {
     if (err) {
       return callback(err);
-    }
-
-    if (compilation.errors.length > 0) {
-      return callback(compilation.errors[0]);
     }
 
     compilation.fileDependencies.forEach((dep) => {
@@ -151,6 +165,10 @@ export function pitch(request) {
     compilation.contextDependencies.forEach((dep) => {
       this.addContextDependency(dep);
     }, this);
+
+    if (compilation.errors.length > 0) {
+      return callback(compilation.errors[0]);
+    }
 
     if (!source) {
       return callback(new Error("Didn't get a result from child compiler"));


rjgotten:
Not quite sure what's up here; I _did_ sign the CLA but it remains pending as not signed?

src/loader.js
```diff
@@ -152,6 +166,10 @@ export function pitch(request) {
       this.addContextDependency(dep);
     }, this);
 
+    if (compilation.errors.length > 0) {
+      return callback(compilation.errors[0]);
+    }
```
evilebottnawi: Why it is moved in other place?

evilebottnawi: For catching errors, right?

rjgotten: Yes. It was moved down passed the dependencies, to ensure those dependenices are always tracked.

Flipping the error state into a success state on the parent compilation, means that it will also update its registered dependencies. (If the parent compilation would fail, it would retain the previous set.)

If we'd drop out of the child compilation because of an error, _before_ the dependencies of the child compilation are moved into the parent, we would 'lose' them. This way, we won't.

rjgotten:
> We need add tests, you can find hmr tests in test directory

I'll have a look for those and will try to figure out how to add a test for this.

rjgotten:
@evilebottnawi 

Had another long look at how to make a test work.

A proper test for the full HMR flow (which in general is indeed currently not actually tested) would, I think, require something more or less like the following:

- A server implementation using `express` with `webpack-dev-middleware` and `webpack-hot-middleware`
- A browser that can be instrumented, e.g. using `puppeteer`
- A basic testcase webpack project, like the one in the repository listed under 'additional info'

An overarching test harness is then needed to close the loop between server and client:
1. Start the Express server (and the webpack compilation)
2. Boot up Puppeteer and load the testcase `index.html` page
3. Start logging console output from Puppeteer.
4. Wait for the "[HMR] Connected" message
5. Change the source Less file on disk. 
6. Wait for the batch of "[HMR]" messages to come in from Puppeteer
7. Match against expected output.
8. etc. etc. following through the entire test scenario.


Does that sound acceptable?

evilebottnawi:
You don't need this stuff, just mock `less-loader` and throw error

rjgotten:
> You don't need this stuff, just mock `less-loader` and throw error, no need e2e tests, unit tests are enough

If you're only interested in the code output containing the expected hot-loader along with the error, yes; _that_ test I can build for you easily. Just a dummy loader which always errors, chained onto the MiniCssExtractPlugin's loader.

evilebottnawi:
@rjgotten this enoguth

src/loader.js
ScriptedAlchemy: Would there be a case where multiple errors need to be shown in the errors array?

rjgotten:
> @evilebottnawi  
> We need add tests, you can find hmr tests in test directory

Added a test which verifies that a module with compilation errors first emits `hot` bindings before throwing its error.

Of special note:
I had to introduce a convention on the `TestCases.test.js` that processes the `/cases/` folder. That convention is that cases starting with a `~` are _expected_ to throw and should not cause the unit test to fail. (As you might expect; the compiler still outputs an error when rendering the adapted HMR error recovery code.)

src/loader.js
rjgotten: @ScriptedAlchemy 
Not sure actually. The current version of the plugin only reports the first child compilation error as well.

test/cases/~hmr-resilience/error-loader.js
```diff
@@ -0,0 +1,4 @@
+module.exports = function loader() {
```
evilebottnawi: Don't use `~` in filename

test/cases/~hmr-resilience/webpack.config.js
```diff
@@ -0,0 +1,62 @@
...
+import Self from '../../../src';
+
+module.exports = {
+  entry: './index.css',
+  mode: 'development',
+  devtool: false,
+  output: {
+    filename: '[name].js',
+  },
+  optimization: {
```
evilebottnawi: Why it is necessary for test?

test/cases/~hmr-resilience/expected/check.js
```diff
@@ -0,0 +1,19 @@
...
+
+/***/ "./index.css":
+/***/ (function(module, exports, __webpack_require__) {
+
+// extracted by mini-css-extract-plugin
+    if(true) {
+      // 1479427200000
+      var cssReload = __webpack_require__("../../../src/hmr/hotModuleReplacement.js")(module.i, {"hmr":true,"locals":false});
+      module.hot.dispose(cssReload);
+      module.hot.accept(function(){});
```
evilebottnawi: Why?

test/cases/~hmr-resilience/error-loader.js
rjgotten: Any other suggestion then for a moniker to use to indicate tests that are expected to fail?

`Testcases.test.js` works by enumerating directory names and running all the cases, then bailing if one of them has a compilation error.  But here we have a testcase which is expected to have a compilation error. First of its kind.

test/cases/~hmr-resilience/webpack.config.js
rjgotten: output.filename is a leftover. I'll remove it.

test/cases/~hmr-resilience/expected/check.js
rjgotten: A little more context please:
why _what_ ?

test/cases/~hmr-resilience/error-loader.js
evilebottnawi: @rjgotten just use `fail` word in test name

test/cases/~hmr-resilience/expected/check.js
evilebottnawi: Why we need this `module.hot.accept(function(){});`? here, please add comment to source doe, because this line is misleading for other developers

rjgotten: Right. Will add a small description in the loader.js where this particular line of code is emitted.

rjgotten: Done

test/cases/~hmr-resilience/webpack.config.js
rjgotten: Removed the filename leftover, and left a comment explaining why the optimization chunk settings are being used.

test/cases/~hmr-resilience/error-loader.js
rjgotten: Adjusted the test pattern in `TestCases.test.js` to `/-fail$/` i.e. any test suffixed with `-fail` is expected to error. And in accordance renamed the testcase to `hmr-resilience-fail`, getting rid of the `~`.

rjgotten:
> @evilebottnawi 
> Please look on notes

Just a friendly ping to inform you that your noted remarks were processed.
Not sure if Github is otherwise notifying you.

evilebottnawi:
in todo list

ron23:
Hi, any progress on this?

jimblue:
I'm having the exact same issue.
Would be awesome to finish this PR.
What still need to be done ?

rjgotten:
> What still need to be done ?

@evilebottnawi has to approve it and merge it.
That's really all there's left to it, afaik.

There's a trivial merge-conflict in `src/loader.js` which is basically a function rename, introduced because this PR has been lingering for half a year. But they can still easily resolve it.

jimblue:
@rjgotten thank your for your message.
I now @evilebottnawi is pretty busy working on many projects.
So I'm not surprised that it's a bit long to merge.
He probably don't have the time or maybe just forget this PR 😄

jimblue:
Hey @evilebottnawi, this is just a friendly ping !
It could be nice to finish this PR, she's getting old 😄 
Thanks

PR #455, opened 2019-10-17
fix(order): find modules from each chunk in group

### LINKED ISSUE #257: "Cannot read property 'pop' of undefined" with a common cache group
I ran into this after upgrading to 0.4.2.

```
ERROR in chunk common [initial]
common.css
Cannot read property 'pop' of undefined
TypeError: Cannot read property 'pop' of undefined
    at MiniCssExtractPlugin.renderContentAsset (.../node_modules/mini-css-extract-plugin/dist/index.js:343:44)
    at Object.render (.../node_modules/mini-css-extract-plugin/dist/index.js:175:32)
    at Compilation.createChunkAssets (.../node_modules/webpack/lib/Compilation.js:2358:29)
    at hooks.optimizeTree.callAsync.err (.../node_modules/webpack/lib/Compilation.js:1268:10)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (.../node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook [as _callAsync] (.../node_modules/tapable/lib/Hook.js:35:21)
    at Compilation.seal (.../node_modules/webpack/lib/Compilation.js:1213:27)
    at hooks.make.callAsync.err (.../node_modules/webpack/lib/Compiler.js:547:17)
    at _done (eval at create (.../node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:9:1)
    at _err16 (eval at create (.../node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:218:22)
    at _addModuleChain (.../node_modules/webpack/lib/Compilation.js:1064:12)
    at processModuleDependencies.err (.../node_modules/webpack/lib/Compilation.js:980:9)
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

common.css is set up like so:
```js
optimization: {
  splitChunks: {
    cacheGroups: {
      common: {
        name: 'common',
        chunks: 'initial',
        minChunks: 3,
        priority: 10,
        reuseExistingChunk: true
      }
    }
  }
},
```

And has an explicit entry with the following:
```js
common: [
  path.resolve(srcPath, 'vendor.scss'),
  path.resolve(srcPath, 'main.scss')
]
```

ryansully:
fixes #257

This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

As discussed in #257, when the plugin calls `modulesByChunkGroup`, it calls the chunk group's `getModuleIndex2()` function from the ChunkGroup object's `_moduleIndices2` Map property. However, the error occurs because the module it's trying to find the index for doesn't exist in `_moduleIndices2`, causing `modulesByChunkGroup` to return an empty array. The for loop that uses the array never iterates, so `bestMatch` is never set and is left undefined, resulting in an error when trying to call `.pop()` on that undefined object.

This fix checks if `getModuleIndex2()` returns no matching modules and, if so, dives into each chunk in the chunk group to find matching modules.

### Breaking Changes
None.

### Additional Info
While this solution dives into a chunk group's chunks to find modules that weren't in the chunk group's `_moduleIndices2` property, it might indicate a bug in webpack where that property is not being fully populated as it should be. If that is the case, then this fix should no longer be necessary, but perhaps it wouldn't hurt to have an extra guard for finding chunk modules.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -421,7 +421,7 @@ class MiniCssExtractPlugin {
       // This loop also gathers dependencies from the ordered lists
       // Lists are in reverse order to allow to use Array.pop()
       const modulesByChunkGroup = Array.from(chunk.groupsIterable, (cg) => {
-        const sortedModules = modules
+        let sortedModules = modules
           .map((m) => {
             return {
               module: m,
@@ -432,6 +432,18 @@ class MiniCssExtractPlugin {
           .filter((item) => item.index !== undefined)
           .sort((a, b) => b.index - a.index)
           .map((item) => item.module);
+
+        // if no modules were found by getModuleIndex2, dive into each chunk
+        // in the group
+        if (!sortedModules || !sortedModules.length) {
+          sortedModules = cg.chunks
+            // reduce each chunk's modules into a flat array
+            .reduce((arr, ch) => [...arr, ...ch.modulesIterable], [])
+            // filter only the modules that match
+            .filter((m) => modules.find((mod) => mod === m))
+            // sort in reverse order
+            .sort((a, b) => a.index2 - b.index2);
+        }
 
         for (let i = 0; i < sortedModules.length; i++) {
           const set = moduleDependencies.get(sortedModules[i]);


ryansully:
I will add tests soon so the coverage delta doesn't change.

ryansully:
I'm afraid writing a test for this will be difficult; running tests locally fails for me and I haven't been able to get them to pass.

(The `lint:prettier` script also fails with an error, so I disabled it just to get Jest to run.)

evilebottnawi:
Just use prettier to fix problem

doried-a-a:
Thanks folks for being working on this. Any updates?

PR #458, opened 2019-10-31
chore: initial work on webpack 5 integration

andriijas:
This PR contains a:

- [x] new **feature**

### Motivation / Use-Case

Webpack 5 is coming, lets work together to make this plugin compatible!

### Breaking Changes

 Not sure, this PR should be considered DRAFT and not ready yet

### Additional Info

Im a webpack amateur. But there is no better way to try to learn than practice, right?

Please provide feedback if this is on the right track and what more thinks to consider.

I would also like to know if there is a way to make the plugin compatible with both webpack 4 and 5.

### Known issues
```
    ERROR in [initial]
    chunk.getChunkMaps is not a function TypeError: chunk.getChunkMaps is not a function
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "bugs": "https://github.com/webpack-contrib/mini-css-extract-plugin/issues",
   "main": "dist/cjs.js",
   "engines": {
-    "node": ">= 6.9.0"
+    "node": ">=10.13.0"
   },
   "scripts": {
     "start": "npm run build -- -w",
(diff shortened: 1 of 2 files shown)


evilebottnawi:
Don't worry we update plugin before stable release webpack@5, right now plugin doesn't work?

andriijas:
@evilebottnawi Thanks! Im not worried :)  There are a couple of deprecation warnings and I would appreciate help and/or feedback to get a beta of this plugin without the warnings for integration in https://github.com/facebook/create-react-app/pull/7929

Thanks 🙌

evilebottnawi:
@andriijas just ignore deprecation messages right now, it is not break code or something important

evilebottnawi:
Compatibility is fixed, release will be soon

PR #459, opened 2019-10-31
Add support for the insert option

giuseppeg:
This PR contains a:

New *feature*.

### Motivation / Use-Case

This new option mirrors the behavior of `style-loader` which allows users to pass a custom `insert` function that takes care of inserting chunks. This is particularly useful when the initiator is in the parent window but the application renders inside of an `iframe`.

### Breaking Changes

None. This change is backwards compatible: when this option is not provided the plugin will keep inserting link tags as before i.e. to the document's `head`.

### Additional Info

I found a similar PR https://github.com/webpack-contrib/mini-css-extract-plugin/pull/371 but the proposed API forces the user to return an element that the plugin will use as insertion point.

### DIFF
--- a/test/cases/insert-function/insert-function.test.js
+++ b/test/cases/insert-function/insert-function.test.js
@@ -1,0 +1,33 @@
+/* globals document, getComputedStyle */
+
+import path from 'path';
+
+import webpack from 'webpack';
+
+import config from './webpack.config';
+
+describe('options.insert as a function', () => {
+  it('inserts the bundle on the page', (done) => {
+    const outputPath = path.resolve(__dirname, 'expected/index.js');
+    webpack({
+      ...config,
+      output: {
+        path: outputPath,
+        libraryTarget: 'umd',
+        library: 'mini',
+      },
+    }).run(() => {
+      let computedValue = getComputedStyle(document.body).backgroundColor;
+      expect(computedValue).toBe('');
+
+      const script = document.createElement('script');
+      script.src = outputPath;
+      document.head.appendChild(script);
+
+      computedValue = getComputedStyle(document.body).backgroundColor;
+      expect(computedValue).toBe('rgba(0, 0, 0, 0)');
+
+      done();
+    });
+  });
+});
--- a/README.md
+++ b/README.md
@@ -343,29 +343,76 @@ module.exports = {
 };
 ```
 
-#### Module Filename Option
+### Plugin Options
+
+#### moduleFilename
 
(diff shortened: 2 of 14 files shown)


giuseppeg:
@evilebottnawi thank you for taking a look. Can you be a bit more specific as for what kind of extra tests we need to merge this PR? I have added some already but I wasn't quite sure if those were enough or the right way to test this feature.

evilebottnawi:
We need test case, here examples https://github.com/webpack-contrib/mini-css-extract-plugin/tree/master/test/cases, just add test for `string` and `Function`

giuseppeg:
Should I just create a new folder in `cases`? If so I thought that those kind of tests are not end to end and don't test runtime (eg. insert a link tag on a page) - please correct me if I am wrong.

evilebottnawi:
> Should I just create a new folder in cases?

yes, `insert-string` and `insert-function`

> If so I thought that those kind of tests are not end to end and don't test runtime (eg. insert a link tag on a page)

Let's implement this, it is not difficult

132yse:
Is there any progress here? It's not very difficult, but it's urgent.

giuseppeg:
@132yse I might have time to work on this sometime next week. If you need it earlier please go ahead and take over it, you'd just need to add tests.

test/cases/insert-function/insert-function.test.js
```diff
@@ -0,0 +1,33 @@
...
+        path: outputPath,
+        libraryTarget: 'umd',
+        library: 'mini',
+      },
+    }).run(() => {
+      let computedValue = getComputedStyle(document.body).backgroundColor;
+      expect(computedValue).toBe('');
+
+      const script = document.createElement('script');
+      script.src = outputPath;
```
giuseppeg: @evilebottnawi it seems that `webpack` is not emitting this file and it 404s. Also I am not sure jsdom supports `getComputedStyle`. In that case I think we either get rid of this test or need to set up something like Puppeteer which IMO is out of scope.

ankurbigcomm:
@giuseppeg @evilebottnawi Any Update here?!

giuseppeg:
I think @ScriptedAlchemy took over this but his PR is blocked.

ScriptedAlchemy:
Merge my PR into your branch if needed. I’m not blocked just ran out of time to perform the last requests. Started working on module Federation and that’s a monster that has taken all my capacity

ScriptedAlchemy:
If you need it now. Extract css has us it for months. https://github.com/faceyspacey/extract-css-chunks-webpack-plugin

evilebottnawi:
https://github.com/webpack-contrib/mini-css-extract-plugin/pull/609 improved

ScriptedAlchemy:
Yaaaay

PR #464, opened 2019-11-15
Add option to export as ES Module

laverdet:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

With this simple option Webpack can entirely remove internal CSS names from all source files. Consider the following simple project:
https://gist.github.com/laverdet/560a89734324c114f4e1646ae6283f2e

If you run without the new option you can find this in `dist/main.js`. Note that "container" is still visible in the generated source file:

`e.exports={container:"RX7MVaJTvfI0d1TLV5iSh"}` ... `function(e,t,n){"use strict";n.r(t);var r=n(0),o=n.n(r);console.log(o.a.container)}`

Running with the new option you'll see Webpack can just inject the classname wherever it's needed. The original "container" classname is no longer visible. With aggressive classname mangling you can drop some bytes.

`function(e,t,r){"use strict";r.r(t);console.log("RX7MVaJTvfI0d1TLV5iSh")}`

### Breaking Changes

None

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -202,7 +202,9 @@ export function pitch(request) {
 
     let resultSource = `// extracted by ${pluginName}`;
     const result = locals
-      ? `\nmodule.exports = ${JSON.stringify(locals)};`
+      ? (options.useExport ?
+         `\nexport default ${JSON.stringify(locals)};` :
+         `\nmodule.exports = ${JSON.stringify(locals)};`)
       : '';
 
     resultSource += options.hmr


laverdet:
I executed the CLA. I also reverted the `package-lock.json` changes because that was just npm being wild. This change is really simple and if you look at the generated code for a typical project it's really a no brainer.

evilebottnawi:
Done in master

laverdet:
Thank you for doing the legwork on landing this with tests and documentation!

PR #475, opened 2019-12-20
feat: new `esModule` option

### LINKED ISSUE #424: Add a configuration flag to enable ES module syntax

evilebottnawi:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #424 fixes #433

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,12 +41,49 @@ Default: the `publicPath` in `webpackOptions.output`
 
 Specifies a custom public path for the target file(s).
 
+#### `esModule`
+
+Type: `Boolean`
+Default: `false`
+
+By default, `mini-css-extract-plugin` generates JS modules that use the CommonJS modules syntax.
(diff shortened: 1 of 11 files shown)


Cryrivers:
should the semantic equivalence to `module.exports = {...}` be `export const XXX = YYY` one-by-one instead of `export default {...}`, the the latter one might be also better for tree-shaking?

PR #482, opened 2019-12-25
fix: improve contenthash generation

### LINKED ISSUE #454: contenthash generates same filename for different contents
- Operating System: macOS 10.14.6
- Node Version: v8.11.2
- NPM Version: 6.0.1
- webpack Version: 4.41.2
- mini-css-extract-plugin Version: 0.8.0

### Expected Behavior

Should generate different filename when content is changed.

### Actual Behavior

Same filename was generated when content is updated.

### How Do We Reproduce?

https://github.com/zlk89/webpack-css-contenthash-diff
- `yarn` to install packages
- `yarn build`
- under `dist` folder, you will find two css files with same content hash, while content is different.

Basically contenthash is same, when import statement order is different:
```
// in js
@import 'a.scss';
@import 'b.scss';
```

```
// in js
@import 'b.scss';
@import 'a.scss';
```

alexander-akait:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #454

### Breaking Changes

No

### Additional Info

/cc @sokra 

Do we need more logic for contenthash generation as in https://github.com/webpack/webpack/blob/master/lib/javascript/JavascriptModulesPlugin.js


src/index.js
```diff
@@ -221,9 +221,21 @@ class MiniCssExtractPlugin {
...
+          hashDigestLength,
+          hashFunction,
+        } = outputOptions;
         const hash = createHash(hashFunction);
 
+        if (hashSalt) {
+          hash.update(hashSalt);
+        }
+
+        hash.update(`${chunk.id} `);
```
sokra: These two are not part of the content of the css file

evilebottnawi: @sokra without it, difference async chunks have same contenthash, but difference content, here example https://github.com/webpack-contrib/mini-css-extract-plugin/issues/454

adriantran: @sokra @evilebottnawi any update on this?

yingzhe: As far as I can tell, CSS modules don't have `id` property set so the hash is going to be the same. Would it be better to add `hash.update(m.index2)` (since index2 is used to sort the modules before rendering content asset) below after `m.updateHash(hash)` ?

PR #486, opened 2019-12-28
feat: Enable `esModule` by default

### LINKED ISSUE #485: Enable esModule by default
- Operating System: N/A
- Node Version: N/A
- NPM Version: N/A
- webpack Version: 4
- mini-css-extract-plugin Version: 0.9.0

### Expected Behavior / Situation
The `esModule` option is enabled by default.

### Actual Behavior / Situation
The `esModule` option is **not** enabled by default.

### Modification Proposal
In #475 a new `esModule` loader option was added, which causes the loader part of `mini-css-extract-plugin` to output ES Module style syntax instead of CJS ([docs](https://github.com/webpack-contrib/mini-css-extract-plugin#esmodule)).

Currently the feature is disabled by default, and to gain the benefits (eg improved tree shaking), people have to enable it explicitly by setting it to `true`.

I propose enabling it by default, so that:
1. People don't have to know about the feature to gain the benefits (not everyone will read about it/understand the benefits)
2. Reduce the amount of webpack config boilerplate when using `mini-css-extract-plugin`

Since webpack 4 already natively supports ES Modules, and `mini-css-extract-plugin` only supports webpack 4 or newer, it seems this could land as a feature change rather than a breaking change, though I'd be fine with either, so long as we can enable it by default :-)

edmorley:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Since it can allow webpack's module concatenation and tree shaking to work more effectively, and having `esModule` disabled by default means more boilerplate and having to know that the feature exists.

### Breaking Changes

None, since webpack 4 supports ES Modules syntax out of the box (and doesn't output ESM itself), and this plugin only supports webpack 4 and newer.

### Additional Info

Fixes #485.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,12 +155,12 @@ module.exports = {
 ### `esModule`
 
 Type: `Boolean`
-Default: `false`
+Default: `true`
 
-By default, `mini-css-extract-plugin` generates JS modules that use the CommonJS modules syntax.
+By default, `mini-css-extract-plugin` generates JS modules that use the ES module syntax.
 There are some cases in which using ES modules is beneficial, like in the case of [module concatenation](https://webpack.js.org/plugins/module-concatenation-plugin/) and [tree shaking](https://webpack.js.org/guides/tree-shaking/).
 
-You can enable a ES module syntax using:
+If you instead require the output to use CommonJS module syntax, disable `esModule` mode like so:
 
 **webpack.config.js**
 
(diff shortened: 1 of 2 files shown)


edmorley:
@evilebottnawi Hi :-) I don't suppose you have a moment to review this?

sokra:
Afaik this also has a functional difference when consuming modules with `require` and therefore must be a breaking change.

evilebottnawi:
Done https://github.com/webpack-contrib/mini-css-extract-plugin/pull/602, release will be today

Issue #492, opened 2020-01-11
Styles should not be inserted into HEAD in Safari


theKashey:
- Operating System: OSX
- Node Version: does not matter
- NPM Version: does not matter
- webpack Version: does not matter
- mini-css-extract-plugin Version: does not matter

### Expected Behavior

Expected that application behaviour would be the same across different browsers.

### Actual Behavior

In `Safari` the whole rendering would be suspended while `style`, added to the `head`, is loading.
Adding `style` to the `body` resolves the problem.

### Code

[This](https://github.com/hkhere/loadable-components_issue-509) example reproduces an issue originally created for [loadable-components](https://github.com/gregberge/loadable-components/issues/509).

It delays serving for `.css` file for 10 second, and all this time nothing is rendered.

### How Do We Reproduce?

1. add `link` to the `head` in Safary
2. render some animation (like counter) in on the page
3. page is expected to be fronzen all time `link` is loading the data

### How to fix?

For Safari "dynamic" styles has to be added into `Body`. It resolves the issue.
- ongoing PR https://github.com/webpack-contrib/mini-css-extract-plugin/pull/459 should fix the problem, however it should be considered as a default behavior.
- I would say that's a WebKit bug, but could not find something at their bug tracker
- see [this PR](https://github.com/gregberge/loadable-components/pull/506#issuecomment-573277582) for a bit more context

ScriptedAlchemy:
@evilebottnawi i could pr it appending to the body instead. Or an option could be added to the plugin like ‘appendTo’ thoughts?

theKashey:
First one has to investigate the consequences. Plus should new `links` be added to the beginning of the `body`, or appended to the `end`?
Look like the issue is here already for a while, let's not rush.

hkhere:
@theKashey It is not just style files but js files too could cause the whole rendering suspended. The dom  is acutally loaded (we can examine it in the devtools). Should this be reported to webkit?

theKashey:
1. Yes. That's a purely webkit bug
2. Also yes... it would take some time to resolve it, as well to release a new version, and ship an update.

I would say - it might take a year to resolve this issue _via_ webkit, so some other solution should be created.

ogonkov:
inserting styles to `<header>` is a common practice, it is definitely shouldn't be changed

theKashey:
That's not how you should think about it. Now "what everybody is doing it", but "why everybody is doing it". And really - there is no reason. 

[...]

PR #495, opened 2020-01-19
feat: Adding insert option

ScriptedAlchemy:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

In Safari the whole rendering would be suspended while style, added to the head, is loading.
Adding style to the body resolves the problem.

This PR gives users the ability to inject async styles into a specified location

### Breaking Changes

none

### Additional Info
Fix: #492 #459


ScriptedAlchemy:
This resolves async issues when loading CSS. Causing WebKit based browsers to block the render until styles are added to the head.

insert option will default to the head, you can choose another location either with a string `body` or a function which is passed into the plugin at compile time

ScriptedAlchemy:
@evilebottnawi - Looks like it failed on commit lint, i did reword it but doesn't appear to fix it in CI. All tests and lints are passing

src/index.js
```diff
@@ -376,8 +382,9 @@ class MiniCssExtractPlugin {
                         '}',
                       ])
                     : '',
-                  'var head = document.getElementsByTagName("head")[0];',
-                  'head.appendChild(linkTag);',
+                  `var insert = ${insert};`,
+                  `if (typeof insert === 'function') { insert(linkTag); }`,
+                  `else { var target = document.querySelector(${insert}); target && insert === 'body' ? target && target.insertBefore(linkTag,target.firstChild) : target.appendChild(linkTag); } `,
```
evilebottnawi: We should avoid extra logic here, if developer what to insert into body they should use `Function`

ScriptedAlchemy: @evilebottnawi so no string option at all? Only function

ScriptedAlchemy: Can amend the PR

evilebottnawi: Yes, only functions

ScriptedAlchemy: updated, please re-review @evilebottnawi

src/index.js
```diff
@@ -383,8 +378,8 @@ class MiniCssExtractPlugin {
                       ])
                     : '',
                   `var insert = ${insert};`,
-                  `if (typeof insert === 'function') { insert(linkTag); }`,
-                  `else { var target = document.querySelector(${insert}); target && insert === 'body' ? target && target.insertBefore(linkTag,target.firstChild) : target.appendChild(linkTag); } `,
+                  `if (insert) { insert(linkTag); }`,
+                  `else {var head = document.getElementsByTagName("head")[0]; head.appendChild(linkTag);} `,
```
evilebottnawi: Can we simplify that, we can generate runtime based on type of option and reduce size of runtime code

ScriptedAlchemy: So pretty much append the correct string of code based on the option a compile time. Rather than logical operators at runtime

ScriptedAlchemy: Done

ScriptedAlchemy:
@evilebottnawi Bumping on this PR

evilebottnawi:
Something broken with CI :confused:

ScriptedAlchemy:
@evilebottnawi  I think its the plugin options

Do you know how i can fix
```ValidationError: Mini CSS Extract Plugin Invalid Options

options.insert should pass "instanceof" keyword validation```

**Comment by ScriptedAlchemy** [comment]
@evilebottnawi found the problem! CI seems to be working now

**Comment by ScriptedAlchemy** [comment]
@evilebottnawi good enough to merge or is there a CI failure I’m causing?

**Comment by evilebottnawi** [comment]
I will look at this today/tomorrow and fix the problem with CI, good work, thanks

File: src/index.js [hunk 3]
```diff
@@ -101,14 +101,15 @@ class CssModuleFactory {
 class MiniCssExtractPlugin {
   constructor(options = {}) {
     validateOptions(schema, options, 'Mini CSS Extract Plugin');
-
+    const insert = options.insert ? options.insert.toString() : null;
```
evilebottnawi: Why do not move that to bottom and use

```js
this.options.insert
 ? Template.asString([this.options.insert, 'insert(linkTag);'])
 : 'var head = document.getElementsByTagName("head")[0]; head.appendChild(linkTag)'
```

And avoid extra code? Because now, `insert` is always `string`

ScriptedAlchemy: Will do

README.md
```diff
@@ -578,6 +578,38 @@ module.exports = {
...
+
+If you target an [iframe](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement) make sure that the parent document has sufficient access rights to reach into the frame document and append elements to it.
+
+#### `insert`
+
+Allows to override default behavior and insert styles at any position.
+
+> ⚠ Do not forget that this code will run in the browser alongside your application. Since not all browsers support latest ECMA features like `let`, `const`, `arrow function expression` and etc we recommend you to use only ECMA 5 features and syntax.
+
+> ⚠ The `insert` function is serialized to string and passed to the plugin. This means that it won't have access to the scope of the webpack configuration module.
```
sokra: I find this a bit weird and think this could be confusing the users.

Doesn't something simple like `insert: "head" | "body"` fulfill the use case?

src/plugin-options.json
```diff
@@ -13,6 +13,10 @@
     },
     "ignoreOrder": {
       "type": "boolean"
+    },
+    "insert": {
+      "description": "Inserts `<link>` at the given position (https://github.com/faceyspacey/extract-css-chunks-webpack-pluginn#insert).",
```
sokra: wrong link

README.md
ScriptedAlchemy: Initially, I supported both use cases. I can reapply that work?

evilebottnawi: hm, maybe `insert: "head" | "querySelector" | Function`, `querySelector` can be `body`/`#head`/`.class`/etc

giuseppeg: @sokra I started to work on this feature because I needed to be able to control the insertion bits, specifically inject inside an iframe.

@evilebottnawi suggestions sounds good to me. Can we merge this PR?

PR #509, opened 2020-03-13
feat(inlining): add {esModule:'named'} option

developit:
This PR contains a:

- [x] new **feature**

### Motivation / Use-Case

Currently, CSS Modules usage in Webpack produces className mappings as JSON modules in bundles. Now that `css-loader` and `mini-css-extract-plugin` both support the `esModule` option, there is a way to "dissolve" these mapping modules entirely. The only required change is that className mappings must be exported as named exports rather than an object default export:

```js
// current:
export default {"a":"a_12345", "b":"b_12345"}
// revised output:
export const a = "a_12345"
export const b = "b_12345"
```

The resulting bundle improvement is ... awesome:

<table><thead><tr><th>Before:</th><th>After:</th></tr></thead>
<tbody><tr><td>

```js
[
  function(e, t, n) {
    e.exports = {
      className1: '_1QXSp-vHE2lSYKI1zPJzRj',
      className2: '_2HdeYqDJ0HpgTdR_Whcvls',
    };
  },
  function(e, t, n) {
    n.r(t);
    var r = n(0);
    document.body.insertAdjacentElement(
      'afterend',
      `<div class="${r.className1}"></div>`,
    ),
      document.body.insertAdjacentElement(
        'afterend',
        `<div class="${r.className2}"></div>`,
      );
  },
]
```

</td><td>

```js
[
  function(e, t, n) {
    n.r(t);
    document.body.insertAdjacentElement(
      'afterend',
      '<div class="_1QXSp-vHE2lSYKI1zPJzRj"></div>',
    ),
      document.body.insertAdjacentElement(
        'afterend',
        '<div class="_2HdeYqDJ0HpgTdR_Whcvls"></div>',
      );
  },
]
```

</td></tr></tbody></table>

### Breaking Changes

This isn't a breaking change, because it is only enabled when a new `"named"` value is passed for the `esModule` option.

### Additional Info

I published [constant-locals-loader](https://www.npmjs.com/package/constant-locals-loader), which performs the conversion of this loader's default exported object into named exports. The output of this PR should be the same as the result of that loader.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -213,7 +213,7 @@ export function pitch(request) {
       result = '';
       if (locals) {
         for (const key in locals) {
-          if (Object.hasOwnProperty.call(locals, key)) {
+          if (Object.prototype.hasOwnProperty.call(locals, key)) {
             if (isInvalidKey.test(key)) {
               skipped.push(key);
             } else {
--- a/src/loader.js
+++ b/src/loader.js
@@ -205,11 +205,38 @@ export function pitch(request) {
 
     const esModule =
       typeof options.esModule !== 'undefined' ? options.esModule : false;
-    const result = locals
-      ? `\n${esModule ? 'export default' : 'module.exports ='} ${JSON.stringify(
-          locals
-        )};`
-      : '';
+
+    let result;
+    if (esModule === 'named') {
+      const skipped = [];
+      const isInvalidKey = /(^[0-9]|-)/;
+      result = '';
+      if (locals) {
+        for (const key in locals) {
+          if (Object.hasOwnProperty.call(locals, key)) {
+            if (isInvalidKey.test(key)) {
+              skipped.push(key);
+            } else {
+              result += `export const ${key} = "${locals[key]}";`;
+            }
+          }
+        }
+      }
+      if (skipped.length) {
+        this.emitWarning(
+          new Error(
+            `Skipped invalid CSS className exports: ${skipped}.\n` +
+              `Please set css-loader's "localsConvention" option to "camelCaseOnly".`
+          )
+        );
+      }
+    } else {
+      result = locals
+        ? `\n${
+            esModule ? 'export default' : 'module.exports ='
+          } ${JSON.stringify(locals)};`
+        : '';
+    }
 
     let resultSource = `// extracted by ${pluginName}`;
 


developit:
I think something is up with CI - tests pass locally.

Bnaya:
I wonder: how it would effect dynamic access to classnames?

SinimaWath:
Hi. @developit

I'm not maintainer of mini-css-extract-plugin, but i see that there are 2 unobvious moment
1) What we need to do with unused styles? Them won't be inlined. Is there some possibility to warn developers about unused styles? 
2) Because of first problem, we have 'const' expression, which needs transpiling by, for example, babel-loader.

developit:
@SinimaWath both good points. The `const` issue could easily be addressed by switching to `var`, there should be no effect on the output. It may need to be marked as `/*#__PURE__*/`.

Overall, I am actually not sure this PR's proposed change is the right way forward. As pointed out on Twitter, there seem to be common de-opt cases that disable module concatenation, which isn't really specific to CSS Modules mappings.

In cases where CSS Modules is configured properly and ModuleConcatenation ends up inlining mappings into their importer, the only change necessary to fully dissolve these mappings is increasing Terser's `compress.passes` to `2`. I've opened a PR proposing that be considered as a default for Webpack's production mode:
https://github.com/webpack-contrib/terser-webpack-plugin/pull/225

src/loader.js
```diff
@@ -205,11 +205,38 @@ export function pitch(request) {
...
-      : '';
+
+    let result;
+    if (esModule === 'named') {
+      const skipped = [];
+      const isInvalidKey = /(^[0-9]|-)/;
+      result = '';
+      if (locals) {
+        for (const key in locals) {
+          if (Object.hasOwnProperty.call(locals, key)) {
```
Pomar81:
```suggestion
          if (Object.prototype.hasOwnProperty.call(locals, key)) {
```

BPScott:
> Overall, I am actually not sure this PR's proposed change is the right way forward in terms of achieving better output. It's useful for those who want to use named imports for their CSS classes, which is not currently the way most folks use CSS Modules in Webpack.

I thiiiink people could use import namespaces to get the same usage effect in the esModule:named world:

Old:
 ```js
// styles is the default export which is an object
import styles from './styles.css';
const className = styles.foo;
```

New:
```
// styles is a module, but it acts the same as the styles object
import * as styles from './styles.css';
const className = styles.foo;
``

evilebottnawi:
We are on this, you can use it using `import styles, * as locals from './style.css'`

evilebottnawi:
Done here https://github.com/webpack-contrib/mini-css-extract-plugin/pull/575, sorry for delay

PR #511, opened 2020-03-17
feat(settings): support chunkFilename as function

yoannmoinet:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Lately we split our webpack build in different processes.

So `mini-css-extract-plugin` was creating chunks named `1.min.css` for different modules in each builds, overwriting the chunk from other builds. It just wasn't configured to add the hash to the name.

We configured it to add the hash to the name using the `chunkFilename` setting.

But, it named our `common.min.css` bundle as a chunk making it `common.<hash>.min.css`.

The issue is that we load this file explicitly in our templates, so we couldn't predict the new name.

A solution was to use a function for `chunkFilename`, just like webpack's configurations. But this plugin didn't support functions for this setting. 

### Breaking Changes

Not breaking.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -199,8 +199,10 @@ class MiniCssExtractPlugin {
       compilation.mainTemplate.hooks.hashForChunk.tap(
         pluginName,
         (hash, chunk) => {
-          const { chunkFilename } = this.options;
-
+          const chunkFilename = this.getChunkFilenameSetting(
+            chunk,
+            chunk.contentHash[MODULE_TYPE]
+          );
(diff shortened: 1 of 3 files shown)


evilebottnawi:
`chunkFilename` is soft deprecated for webpack@5

yoannmoinet:
What would be the correct process in webpack@5 then?

evilebottnawi:
Supported (only webpack@5)

PR #514, opened 2020-03-22
feat: new hook miniCssExtractPluginBeforeLinkAppend

wangzongxu:
Allow the link tag to be added to the head before doing some extra processing, such as I want to set a load timeout listener

This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

As the link loads, I want to add code that controls the timeout

### Breaking Changes

No

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -13,6 +13,9 @@ const {
   Template,
   util: { createHash },
 } = webpack;
+const {
+  SyncWaterfallHook,
+} = require("tapable");
 
 const MODULE_TYPE = 'css/mini-extract';
 
@@ -129,6 +132,8 @@ class MiniCssExtractPlugin {
 
   apply(compiler) {
     compiler.hooks.thisCompilation.tap(pluginName, (compilation) => {
+      compilation.hooks.miniCssExtractPluginBeforeLinkAppend = new SyncWaterfallHook(['source', 'chunk', 'hash'])
+
       compilation.dependencyFactories.set(
         CssDependency,
         new CssModuleFactory()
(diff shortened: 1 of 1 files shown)


ankurbigcomm:
@wangzongxu Any update on this. This might be critical for one of our applications as well.

wangzongxu:
Sorry

test/BeforeLinkAppendHook.test.js
```diff
@@ -0,0 +1,49 @@
...
+      compiler.hooks.thisCompilation.tap(
+        'RewriteErrorHandler',
+        (compilation) => {
+          compilation.hooks.miniCssExtractPluginBeforeLinkAppend.tap(
+            'RewriteErrorHandler',
+            (source, chunk, hash) => {
+              expect(source).toBeDefined();
+              expect(chunk).toBeDefined();
+              expect(hash).toBeDefined();
+              return 'head.onerror = console.error;';
```
evilebottnawi, line 18: Why do you need rewrite error handle? What is real use case this hook?

wangzongxu: The actual use case is that I need to listen for network error events and initiate a retry，Domain names may be switched during the process.

evilebottnawi: > The actual use case is that I need to listen for network error events and initiate a retry

wangzongxu: I can get a normal error, but I want to set a timeout for the resource load, then switch the domain name and reload it.
If I don't pass the hook, It will be difficult to get the start loading time of the resource.

evilebottnawi: What is the weird use case? Why do you need this? It is potential break cache

Issue #444, opened 2019-09-03
HMR: inferring script source from moduleId is not reliable


jsnajdr:
The HMR code needs to do some magic to convert the `moduleId` of the updated module to the URLs of the CSS chunks that contain the module. These URLs then need to be reloaded.

Trying to enable it in [Calypso](https://github.com/automattic/wp-calypso) reveals at least three bugs:

**Doesn't work with async-loaded chunks**
`getCurrentScriptUrl` relies on `document.currentScript`, which works when the chunk is loaded with `script` tag, but not when a chunk is loaded asynchronously. Then the module is evaluated in a `Promise.then` handler and `document.currentScript` is `null`. No hot reload.

**Is not reliable with shared chunks**
With shared async chunks optimized by `SplitChunksPlugin`, it commonly happens that a module code is duplicated in multiple chunks. To reliably hot-reload their CSS, we need to reload all the CSS chunks where the module is bundled. That doesn't happen at this moment. Only the chunk that was used to actually execute the module is reloaded.

**Fallback to "last `script` tag" almost never works**
When `getCurrentScriptUrl` can't find `document.currentScript`, it falls back to using the `src` attribute of the last `script` tag in the document. That works for very simple apps that load a single JS bundle, but fails in cases like:
- there are shared chunks that together form a chunk group. There's not reason why the last chunk should contain the updated `moduleId`
- the last `script` tag is not owned by webpack, but loads something else. For example, analytics from a 3rd party source
- the last `script` doesn't have a `src` attribute, but is an inline script. Then the HMR code will crash with `TypeError: Cannot read property 'some' of null`

evilebottnawi:
We recommend disable `SplitChunksPlugin` when you want to use `HMR`, it is very hard to implement

evilebottnawi:
And maybe won't fix

evilebottnawi:
Also please provide minimum reproducible test cases

jsnajdr:
> We recommend disable SplitChunksPlugin when you want to use HMR, it is very hard to implement

That still leaves two of three bugs: async loading and `Promise.then`, and unreliable last `script` tag.

Could you shed some light on why the "last `script` tag" code is there at all? It's very easy to break -- why not just look at `document.currentScript` and reject the hot update if it's `null`?

evilebottnawi:
Maybe it is bug, sometimes bugs happen, you can send a PR with fixes

jsnajdr:
[...]

PR #531, opened 2020-05-22
fix: js exception when reloading async chunk (#444)

Rulexec:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

HMR throws exception every time when I do edits in my css files:

```
Uncaught TypeError: Cannot read property 'some' of null
    at getReloadUrl (hotModuleReplacement.js:129)
    at hotModuleReplacement.js:145
    at NodeList.forEach (<anonymous>)
    at reloadStyle (hotModuleReplacement.js:140)
    at update (hotModuleReplacement.js:194)
    at functionCall (hotModuleReplacement.js:24)
```

This fix ignores such obviously wrong updates and allows to use `options.reloadAll`, because this exception happens just before `reloadAll()` call.

Issue — #444 

### Breaking Changes

None.


evilebottnawi:
Or please create reproducible test repo

Rulexec:
Reproducible [test branch](https://github.com/Rulexec/mini-css-extract-plugin/tree/hmr-exception-demo).

I will try to write test, but it can take a while to investigate, how to write such test.

src/hmr/hotModuleReplacement.js
```diff
@@ -227,3 +231,8 @@ module.exports = function(moduleId, options) {
 
   return debounce(update, 50);
 };
+
+// Used from tests
+module.exports.clearCache = function() {
+  srcByModuleId = Object.create(null);
+};
```
evilebottnawi, line 238: We should use mock for test it, no need it in runtime code

Rulexec: How to mock variable inside module?

Currently value inside `srcByModuleId = Object.create(null)` persists between tests and my situation happens only if it's clear.

Issue #512, opened 2020-03-18
Non-deterministic contenthash generation


bschmeck:
We run a Rails app that is deployed to multiple webservers.  During our deploy, we build packs on each machine from the same deployed code.  We have very recently encountered an issue where this process has resulted in packs with different `contenthash` fingerprints in their filename.  There are only 2 CSS files in the chunk, and it seems to be an issue of the order in which those modules are hashed when `contenthash` is being calculated.

- Operating System: CentOS 8
- Node Version: 10.19.0
- NPM Version: 6.13.4
- webpack Version: 4.41.5
- mini-css-extract-plugin Version: 0.8.2

### Expected Behavior

Building packs for the same code on multiple machines should generate identical `contenthash` filenames.

### Actual Behavior

The modules associated with a chunk are hashed in a non-deterministic way, resulting in different fingerprints on different machines.

### How Do We Reproduce?

I'm still working on a reproducable example, but the issue appears to be the order in which `chunk.modulesIterable` returns modules here: https://github.com/webpack-contrib/mini-css-extract-plugin/blob/1ffc393a2e377fe0cc341cfcbc5396e07a8e4077/src/index.js#L227-L231

`modulesIterable` is a `SortableSet`, but it appears that the code needs to explicitly call `sort()` on the set to retrieve values in order.  Adding a call to `chunk.modulesIterable.sort()` before hashing the modules appears to be sufficient to generate a consistent fingerprint.  If that seems to be a reasonable approach, I'm happy to open a PR but am struggling to put together a test for the change.

evilebottnawi:
We known about non-deterministic contenthash problems due source maps, please create reproducible test repo we need investigate, maybe you faced with a new problem

yingzhe:
Left a comment here: https://github.com/webpack-contrib/mini-css-extract-plugin/pull/482
I think the problem is that when generating the `contenthash` generation is based on the order of modules inserted into `chunk.modulesIterable`, but when rendering content asset, the modules are first sorted by `index2`. If `contenthash` generation includes `index2`, maybe that would solve the issue?

evilebottnawi:
Solved for webpack@5 with and without source maps, for webpack@4 solved only without source maps (architecture problems), sorry for delay

PR #533, opened 2020-05-28
fix: contextify sourceMap

helloyou2012:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

some loader (like postcss-loader) generate the source map with the absolute path. Cause contentHash differences on different project paths.


helloyou2012:
#512

helloyou2012:
Ref: https://github.com/webpack/webpack/blob/b7f08b42b85f89e52facac426251d26ec64f2e6d/lib/NormalModule.js#L528

helloyou2012:
I don't think it should be done on css-loader side. Even if do it on css-loader side, it does not guarantee the other loaders contextify. So I think it should do on this plugin. Webpack is also contextify source map first and don't need to absolute sources.

evilebottnawi:
No, we need absolute sources for webpack, you break source maps, other CSS files on a page can have source maps too with relative sources (for example `src/style.css`)

> some loader (like postcss-loader) generate the source map with the absolute path.

postcss-loader doesn't create source maps with absolute paths

helloyou2012:
https://github.com/postcss/postcss-loader/blob/18c93f16164b875861e281c63342bd2f46ba0bf3/src/index.js#L159

evilebottnawi:
@helloyou2012 It is a bug in `postcss-loader`, using `path.resolve` is a wrong idea, without absolute sources webpack will be generate broken source maps

helloyou2012:
path.resolve return absoluted paths.

evilebottnawi:
@helloyou2012 yep, but they may not necessarily be relatively to current process cwd

helloyou2012:
So postcss-loader should keep the original relative path and no need resolve it?

evilebottnawi:
@helloyou2012 postcss-loader should generate right sources with absolute paths, css-loader should contextify using `webpack://`, webpack automatically absolute them https://github.com/webpack/webpack/blob/master/lib/SourceMapDevToolPlugin.js#L78

helloyou2012:
why not contextify using webapck:// on this plugin?

evilebottnawi:
@helloyou2012 because this plugins should not do it

evilebottnawi:
@helloyou2012 and we still have the same problem with `style-loader`, so it is better to fix it on `css-loader` side

helloyou2012:
Is this pr to fix this bug?
https://github.com/webpack-contrib/css-loader/pull/1031

evilebottnawi:
@helloyou2012 yep, bug we need a new a PR, `contextify` is better solution

evilebottnawi:
Fixed on `css-loader` side (now we always pass `webpack://./rootDirectory/my-style.css` sources), fix for `postcss-loader` will be in near future

evilebottnawi:
Just update `css-loader` to latest version

PR #586, opened 2020-09-11
feat: support for multiple inlined stylesheets in the same style tag

wille:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I have multiple CSS chunks in my web application and in some cases I want to inline more than one of these chunks into a single <style> tag.

Currently `data-href` does only support one href so I would have to create one style tag per chunk that is inlined which is not super convenient in my current stack.

```
<style data-href="/static/test.css,static/test2.css">
```


lsycxyj:
I suppose maybe you should do this by html plugin

PR #594, opened 2020-09-28
add types for ts support and support webpack v5 config

YanYuanFE:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

add types for ts support and support webpack v5 config


evilebottnawi (changes requested):
Sorry, we should use `ts` docs to generate types, also it was implemented after webpack@5 stable release

PodaruDragos:
@alexander-akait are there any plans to add `types` directly in this package and not rely on `@types/mini-css-extract-plugin` ?

PR #599, opened 2020-10-07
feat(webpack-5): opt-in to transitive only side effects

sokra:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

mini-css leaves empty "// extract by ..." js modules in the bundle. They are unnecessary.
webpack 5 will analyse the source code as side-effect-free, but as it contains a custom dependency (`CssDependency`), this dependency also need to opt-in into being side-effect-free. But actual it's not completely side-effect-free as it points to CssModules that need to be included in the bundle. So this PR opts in into a "transitive only" side-effects mode, which allows to drop the js module but keeps the dependencies in the bundle.

### Breaking Changes

None

### DIFF
--- a/src/CssDependency.js
+++ b/src/CssDependency.js
@@ -20,6 +20,11 @@ class CssDependency extends webpack.Dependency {
     return `css-module-${this.identifier}-${this.identifierIndex}`;
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  getModuleEvaluationSideEffectsState() {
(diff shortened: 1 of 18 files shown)


evilebottnawi (approved):

l5oo00:
This feature may make the effect of `optimization.chunkIds: "deterministic"` not as expected.

The example repository: https://github.com/l5oo00/demo-for-mini-css-extract-plugin

PR #605, opened 2020-10-09
Docs: fix ambiguous docs

### LINKED ISSUE #516: Usage with style loader - ambiguous docs
Documentation Is:

- [ ] Missing
- [ ] Needed
- [x] Confusing
- [ ] Not Sure?

### Please Explain in Detail...
The documentation says: 

> This plugin should be used only on production builds without style-loader in the loaders chain, especially if you want to have HMR in development

The first part:

> This plugin should be used only on production builds without style-loader in the loaders chain

seems to suggest that this plugin should not be used in development builds, only in production builds and that too without style-loader. Is that understanding correct?

The second part:
`especially if you want to have HMR in development`
seems to suggest that the plugin can be used in development and can also enable HMR.
But if it can be used in development, is style-loader still a problem? Why does it say especially and talk about development, while the previous statement is about production?
This is quite confusing.

### Your Proposal for Changes
If the plugin can also be used in development, and if should always be used without style loader, then the first part may be better written as:
This plugin should be used only on builds without style-loader in the loaders chain

cap-Bernardito:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

Fixes #516 

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,8 @@ Works like [`output.filename`](https://webpack.js.org/configuration/output/#outp
 Type: `String|Function`
 Default: `based on filename`
 
+> i Specifying `chunkFilename` as a `function` is only available in webpack@5
+
 This option determines the name of non-entry chunk files.
 
 Works like [`output.chunkFilename`](https://webpack.js.org/configuration/output/#outputchunkfilename)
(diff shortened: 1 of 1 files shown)


evilebottnawi (approved):

evilebottnawi:
Thanks!

PR #652, opened 2020-11-06
fix: schema

evilebottnawi:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fix schema

### Breaking Changes

Potential yes, but we should force it because a lot of developers do mistakes

### Additional Info

No

### DIFF
--- a/src/loader-options.json
+++ b/src/loader-options.json
@@ -1,6 +1,6 @@
 {
   "type": "object",
-  "additionalProperties": true,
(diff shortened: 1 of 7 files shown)


Shaharif14:
Yes

PR #673, opened 2020-12-22
feat(options): added query option that gets appended to the chunk href

jacob-ebey:
Added query option that gets appended to the chunk href. This option can be a string or a
synchronous function.

This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

It is fairly common to append a query string to assets hosted on platforms like github-pages for cache invalidation. This change allows frameworks that pre-render HTML to append a query string while not loading duplicate chunks on the client.

Check the network log here for the problem: https://jacob-ebey.js.org/preact-webpack-prerender/

### Breaking Changes

N/A

### DIFF
--- a/src/CssLoadingRuntimeModule.js
+++ b/src/CssLoadingRuntimeModule.js
@@ -50,6 +50,13 @@ module.exports = class CssLoadingRuntimeModule extends RuntimeModule {
 
     if (!withLoading && !withHmr) return null;
 
+    let queryPostfix = '';
+    if (typeof this.runtimeOptions.query === 'string') {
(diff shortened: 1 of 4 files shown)


alexander-akait (changes requested):
it is out of scope this plugin, you have the `filename` option https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/plugin-options.json#L5 for this

jacob-ebey:
🤦that totally covers the use-case

PR #681, opened 2021-01-26
fix(contenthash): do not generate mini-css contenthash for chunks without css modules

### LINKED ISSUE #680: Produces bloated runtime in webpack 5 when using `contenthash` or `chunkhash`
- Operating System: OSX 10.15.7
- Node Version: 12.20.1
- NPM Version: 6.14.10
- webpack Version: 5.17.0
- mini-css-extract-plugin Version: 1.3.4

### Expected Behavior

When using `filename: '[name].[contenthash].css'`, it produces a runtime chunk that only contains values for existing CSS files.

### Actual Behavior

Runtime is bloated with replacement maps that point to inexistent CSS files. For example, using this config:

```
  plugins: [new MiniCssExtractPlugin({
    filename: '[name].[id].[contenthash].css',
    chunkFilename: '[name].[id].[contenthash].css',
  })],
```

I got the following in the runtime:

```
/******/ __webpack_require__.miniCssF = (chunkId) => {
/******/ 	// return url for filenames based on template
/******/ 	return "" + ({"179":"main","303":"runtime~main"}[chunkId] || chunkId) + "." + chunkId + "." + {"179":"c539139b868f46f81e90","229":"31d6cfe0d16ae931b73c","851":"31d6cfe0d16ae931b73c"}[chunkId] + ".css";
/******/ };
```

However, only the file `main.179.c539139b868f46f81e90.css` exists (that is correct). The replacement function includes mapping for chunks that don't have CSS (eg: `"303":"runtime~main"` or `"229":"31d6cfe0d16ae931b73c"`). In the case of a big app this results in a bloated runtime.

Note that using `chunkhash` instead of `contenthash` does not solve the issue.

### How Do We Reproduce?

See https://github.com/scinos/mini-css-bug-contenthash for a reproduction repo.

scinos:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Fixes #680

### Breaking Changes

None

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -268,17 +268,20 @@ class MiniCssExtractPlugin {
         const { hashFunction, hashDigest, hashDigestLength } = outputOptions;
         const hash = createHash(hashFunction);
 
-        for (const m of this.getChunkModules(chunk, chunkGraph)) {
-          if (m.type === MODULE_TYPE) {
+        const modules = Array.from(
+          this.getChunkModules(chunk, chunkGraph)
+        ).filter((module) => module.type === MODULE_TYPE);
+        if (modules.length) {
+          for (const m of modules) {
             m.updateHash(hash, { chunkGraph });
           }
+
+          const { contentHash } = chunk;
+
+          contentHash[MODULE_TYPE] = hash
+            .digest(hashDigest)
+            .substring(0, hashDigestLength);
         }
-
-        const { contentHash } = chunk;
-
-        contentHash[MODULE_TYPE] = hash
-          .digest(hashDigest)
-          .substring(0, hashDigestLength);
       });
 
       const { mainTemplate } = compilation;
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -268,17 +268,20 @@ class MiniCssExtractPlugin {
         const { hashFunction, hashDigest, hashDigestLength } = outputOptions;
         const hash = createHash(hashFunction);
 
-        for (const m of this.getChunkModules(chunk, chunkGraph)) {
-          if (m.type === MODULE_TYPE) {
+        const modules = Array.from(
+          this.getChunkModules(chunk, chunkGraph)
+        ).filter((module) => module.type === MODULE_TYPE);
+        if (modules.length) {
```
alexander-akait, line 274: Let's use `continue`, it is faster, just put `// eslint-ignore` if linter will report about it (we need disable this rule, because we have more problems with it)

alexander-akait:
Thanks for the PR, I solved it with better perf, also found place there we can reduce runtime more, so it is only first place

PR #720, opened 2021-03-16
fix: include matchResource in entry request

jquense:
Sorry @alexander-akait, I'm always creating a bunch of work for you for super niche features. I think this should be clear but happy to explain it more if you want.

This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**


jquense:
arg, why is trying to upgrade a single dependency breaking everything!

alexander-akait:
hm, I don't think it is good idea to do in pitch phase, we can faced with infinity loop...

jquense:
you'd certainly know better than me. It _seemed_ ok to me since the original request included it. What's happening is that the original module (with `matchResource`) is sort of being ignored, since mini-extract and style-loader force a new module to be created, this time without the matchResource (since it's not in the request anymore)

I had also thought that the `!!./loaders` might prevent a loop since there is no loader matching?

alexander-akait:
I don't think we should touch `matchResource` in pitch mode here, I think you need do on own side using `pitch`, so I want to look how you do it right now

jquense:
the simplest case is the base4-loader. which doesn't work with style-loader or mini-extract.  

The way more complicate case is here, still a WIP.

https://github.com/4Catalyzer/astroturf/blob/f723dc2d23ece81417f34daf1d6562dba6f1fc15/src/inline-loader.ts

FYI this does seem to work if i manually set `matchResource` on the module, but that seems like a bad idea?

Issue #713, opened 2021-03-06
New option: EmitFile, similar to file-loader to allow for the plugin ***not*** to emit the assets


ferdinando-ferreira:
- Operating System: Any
- Node Version: Any
- NPM Version: Any
- webpack Version: Any
- mini-css-extract-plugin Version: Any

### Feature Proposal
Addition of the option emitFile, similar to the one existing in other plugins like file-loader (see: [https://webpack.js.org/loaders/file-loader/#emitfile](https://webpack.js.org/loaders/file-loader/#emitfile)

### Feature Use Case
Much like in the file-loader use case, it allows for the resource to be extracted from the javascript (reducing its size) without actually creating the resource on disk, useful to when a SSR version of the javascript file is being generated

ferdinando-ferreira:
Would also satisfy #74, an older version of this same request that was abandoned back

ferdinando-ferreira:
> You can just ignore CSS files using `alias.resolve` for SSR build

I would need to study that alternative to ensure it is accurate but, for the sake of continuity I'll just assume it is correct. With that said I don't think it solves the use case being put forward here.

The idea with the proposed functionality (in a similar way with the way it worked for the file-loader [equivalent option](https://webpack.js.org/loaders/file-loader/#emitfile)) is:

1. To ensure that there is exactly **zero** difference (with regards to the extraction of the styles) between the output javascript files for the regular version and the SSR version
2. While at the same time ensuring the only necessary configuration on webpack.config (also with regards to the extraction of the styles) happens to be

```
		plugins: [
			new MiniCssExtractPlugin({
				filename: '[name].css',
				emitFile: isSsr ? false : true
			})
		],
```
If necessary for the PR to be considered I can construct the use case where `alias.resolve` would be insufficient but `emitFile: false` would solve the problem, and for that all I need is assurance there is no other objection to adding the option.

Should I proceed pleading the case or there is a bigger issue here preventing the new option from being considered?

alexander-akait:
> If necessary for the PR to be considered I can construct the use case where alias.resolve would be insufficient but emitFile: false would solve the problem, and for that all I need is assurance there is no other objection to adding the option.

Please provide case

ferdinando-ferreira:
> Please provide case

An usecase was prepared at [ferdinando-ferreira/mini-css-extract-plugin](https://github.com/ferdinando-ferreira/mini-css-extract-plugin)

Instructions to test

```
[...]

Issue #74, opened 2018-03-30
[Feature Request] Option to prevent the actual file from being emitted


ferdinando-ferreira:
Thanks for the excellent module.

Is it possible to have an option similar to file-loader [emitFile](https://webpack.js.org/loaders/file-loader/#emitfile)? The use case is the generation of the server side versions of packages, same use case as the one in the file-loader and the effect is to prevent the actual generation of the file (as it won't actually be used).

Best regards

ferdinando-ferreira:
For the moment I ended up using the plugin located at [Ignore Emit Webpack plugin
](/mrbar42/ignore-emit-webpack-plugin), sinking all the css emitted after the build phase.
Would be nice to have it as a configuration in this plugin but that solves the problem fine so I'm closing the issue

pravdomil:
@ferdinando-ferreira can you please reopen the issue? because I missing `emitFile` option in this package

PR #722, opened 2021-03-17
feat: emitFile option

ferdinando-ferreira:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

Add emit option (default: true).

 * If true, emits a file (writes a file to the filesystem).
 * If false, the plugin will extract the CSS but will not emit the file.

The use case for this this option is the same of the already existing on `file-loader` (see https://github.com/webpack-contrib/file-loader#emitFile): to disable the actual emission of the file, useful for server-side packages

### Breaking Changes

 * n/a
 
### Additional Info
In order to test an use case demonstration was prepared.

```
git clone https://github.com/ferdinando-ferreira/mini-css-extract-plugin --single-branch mini-css-extract-plugin
cd mini-css-extract-plugin
npm install
cd usecase
npm install
npm run build
```

It generates 3 compilations for the same source, with different parameters

**dist/current-plugin**: uses the current plugin as is. Generates the `dist/ssr/index.css`, which is what the creation of the new option is trying to suppress, as the file is useless for SSR
```
 - nonssr:
     - 1042K  index.css
     -  249K   index.js
 - ssr:
     - 1042K  index.css
     - 1666K  index.js

```

**dist/intended-behaviour**: uses the proposed new option `emit`. Does **not** generate the `dist/ssr/index.css`, which is exactly what the new option is set to achieve
```
 - nonssr:
     - 1042K  index.css
     - 249K   index.js
 - ssr:
     - 1666K  index.js
```
**dist/proposed-alternative-1**: uses the alternative proposed by @alexander-akait in https://github.com/webpack-contrib/mini-css-extract-plugin/issues/713#issuecomment-798767784 (_"just don't use this plugin for SSR"_) which, while not generating the `dist/ssr/index.css`, it does that by leaving it unextracted from the `index.js`, which fails to achieve the goal.
```
 - nonssr:
     - 1042K  index.css
     - 249K   index.js
 - ssr:
     - 2711K  index.js
```

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -199,14 +199,6 @@ export function pitch(request) {
       }
 
       addDependencies = (dependencies) => {
-        if (!Array.isArray(dependencies) && dependencies != null) {
-          throw new Error(
-            `Exported value was not extracted as an array: ${JSON.stringify(
-              dependencies
-            )}`
-          );
-        }
-
         const identifierCountMap = new Map();
 
         let lastDep;
--- a/src/loader.js
+++ b/src/loader.js
@@ -191,39 +191,38 @@ export function pitch(request) {
     const assetsInfo = new Map();
     const emit = typeof options.emit !== 'undefined' ? options.emit : true;
 
-    let addDependencies;
     if (emit) {
       for (const asset of compilation.getAssets()) {
         assets[asset.name] = asset.source;
         assetsInfo.set(asset.name, asset.info);
       }
-
-      addDependencies = (dependencies) => {
-        const identifierCountMap = new Map();
-
-        let lastDep;
-
-        for (const dependency of dependencies) {
-          if (!dependency.identifier) {
-            // eslint-disable-next-line no-continue
-            continue;
-          }
-
-          const count = identifierCountMap.get(dependency.identifier) || 0;
-          const CssDependency = MiniCssExtractPlugin.getCssDependency(webpack);
-
-          this._module.addDependency(
-            (lastDep = new CssDependency(dependency, dependency.context, count))
-          );
-          identifierCountMap.set(dependency.identifier, count + 1);
-        }
-
-        if (lastDep) {
-          lastDep.assets = assets;
-          lastDep.assetsInfo = assetsInfo;
-        }
-      };
     }
+
+    const addDependencies = (dependencies) => {
+      const identifierCountMap = new Map();
+
+      let lastDep;
+
+      for (const dependency of dependencies) {
+        if (!dependency.identifier) {
+          // eslint-disable-next-line no-continue
+          continue;
+        }
+
+        const count = identifierCountMap.get(dependency.identifier) || 0;
+        const CssDependency = MiniCssExtractPlugin.getCssDependency(webpack);
+
+        this._module.addDependency(
+          (lastDep = new CssDependency(dependency, dependency.context, count))
+        );
+        identifierCountMap.set(dependency.identifier, count + 1);
+      }
+
+      if (lastDep) {
+        lastDep.assets = assets;
+        lastDep.assetsInfo = assetsInfo;
+      }
+    };
 
     if (error) {
       return callback(error);
@@ -274,30 +273,29 @@ export function pitch(request) {
         locals = exports && exports.locals;
       }
 
-      if (emit) {
-        let dependencies;
+      let dependencies;
 
-        if (!Array.isArray(exports)) {
-          dependencies = [[null, exports]];
-        } else {
-          dependencies = exports.map(([id, content, media, sourceMap]) => {
-            const module = findModuleById(compilation, id);
+      if (!Array.isArray(exports)) {
+        dependencies = [[null, exports]];
+      } else if (!emit) {
+        dependencies = exports;
+      } else {
+        dependencies = exports.map(([id, content, media, sourceMap]) => {
+          const module = findModuleById(compilation, id);
 
-            return {
-              identifier: module.identifier(),
-              context: module.context,
-              content: Buffer.from(content),
-              media,
-              sourceMap: sourceMap
-                ? Buffer.from(JSON.stringify(sourceMap))
-                : // eslint-disable-next-line no-undefined
-                  undefined,
-            };
-          });
-        }
-
-        addDependencies(dependencies);
+          return {
+            identifier: module.identifier(),
+            context: module.context,
+            content: Buffer.from(content),
+            media,
+            sourceMap: sourceMap
+              ? Buffer.from(JSON.stringify(sourceMap))
+              : // eslint-disable-next-line no-undefined
+                undefined,
+          };
+        });
       }
+      addDependencies(dependencies);
     } catch (e) {
       return callback(e);
     }
--- a/src/loader.js
+++ b/src/loader.js
@@ -295,6 +295,7 @@ export function pitch(request) {
           };
         });
       }
+
       addDependencies(dependencies);
     } catch (e) {
       return callback(e);
--- a/src/loader.js
+++ b/src/loader.js
@@ -189,45 +189,49 @@ export function pitch(request) {
   childCompiler.runAsChild((error, entries, compilation) => {
     const assets = Object.create(null);
     const assetsInfo = new Map();
+    const emit = typeof options.emit !== 'undefined' ? options.emit : true;
 
-    for (const asset of compilation.getAssets()) {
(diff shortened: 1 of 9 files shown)


alexander-akait:
Why you need this plugin for SSR?

ferdinando-ferreira:
> Why you need this plugin for SSR?

It is explained in the PR. Without the plugin the resulting ssr `index.js` contains the unextracted css.

alexander-akait:
Sorry, again, just ignore CSS `require`/`import` for SSR

ferdinando-ferreira:
> Sorry, again, just ignore CSS `require`/`import` for SSR

That is also explained in the PR provided usecase and also not possible. The css is not "required" or "imported" but extracted from the .vue single file component by vue-loader and passed to this plugin for extraction.

alexander-akait:
In this case you need to open an issue in `vue-loader`

ferdinando-ferreira:
> In this case you need to open an issue in `vue-loader`

Vue loader delegates css extraction (see https://vue-loader.vuejs.org/guide/extract-css.html#webpack-4) to this plugin.

Extracting css is the main and only reason this plugin exists. [file-loader](https://github.com/webpack-contrib/file-loader), the equivalent for the other file types implements it already.

If there is no technical reason why the resistance to implement it?

alexander-akait:
Because we don't need this option:
- a lot of developers use SSR and don't need this options (for many years no one issue about it)
- you should not use this plugin for SSR, because it does not make sense
- avoid emitting assets is not safe, we update hashes but do not emit nothing, it is bad for caching

alexander-akait:
Also implementation is invalid, because we don't need process something in loader, we can just return empty content

ferdinando-ferreira:
> a lot of developers use SSR and don't need this options 

Because they would just sink the files using `ignore-emit-webpack-plugin`

> (for many years no one issue about it)

#74, nearing 3 years now and that had other parties interested, not just myself.

> you should not use this plugin for SSR, because it does not make sense

It does make sense. If you don't extract the CSS it remains in the .js file. It makes index.js (in this example) a 2.7M files instead of a 1.6M. 

> avoid emitting assets is not safe, we update hashes but do not emit nothing, it is bad for caching

This is not a technical issue. The new option is optional and does not change the default behavior, be it for cache or otherwise. A warning in the doc saying `passing false to this option may be bad for caching` (whatever that means) would be enough.

alexander-akait:
> Because they would just sink the files using ignore-emit-webpack-plugin

Can you provide link on docs or examples?

> #74, nearing 3 years now and that had other parties interested, not just myself.

This issue from you

> It does make sense. If you don't extract the CSS it remains in the .js file. It makes index.js (in this example) a 2.7M files instead of a 1.6M.

But you can just ignore all `require`/`import` for CSS

> This is not a technical issue. The new option is optional and does not change the default behavior, be it for cache or otherwise. A warning in the doc saying passing false to this option may be bad for caching (whatever that means) would be enough.

We don't want to provide unsafe options

alexander-akait:
> Can you provide link on docs or examples?

I mean for SSR and CSS, no need to link on `file-loader`, there other logic

ferdinando-ferreira:
> Why you need this plugin for SSR? (...) you can just ignore all `require`/`import` for CSS

You are correct that the option is not really *required* to be in this plugin, one can just use `null-loader`

```javascript
const webpack = require('webpack');
const path = require('path');
const { VueLoaderPlugin } = require('vue-loader');
// const MiniCssExtractPlugin = require('../../../dist/cjs.js');
module.exports = {
  mode: 'development',
  target: 'node',
  output: {
    path: path.join(__dirname, '..', '..', 'dist', 'current-plugin', 'ssr'),
    publicPath: '/assets'
  },
  entry: {
    index: './src/index.js',
  },
  module: {
    rules: [
      {
        test: /\.vue$/,
        use: 'vue-loader'
      },
      {
        test: /\.css$/,
        use: [
          { loader: "vue-style-loader" },
          { loader: "null-loader" },      // <------------- add this instead
//        { loader: MiniCssExtractPlugin.loader, options: { esModule: false } },
//        { loader: "css-loader" }
        ]
      }
    ]
  },
  plugins: [
     new webpack.DefinePlugin({
      COMPILE_TARGET: JSON.stringify('node')
    }),
    new VueLoaderPlugin(),
//  new MiniCssExtractPlugin({
//    filename: '[name].css'
//  })
  ],
  devtool: false
}
```

It would be nice to have it tho, meaning the same plugin / loader could be used in either case with a single option being the difference.

> Also implementation is invalid, because we don't need process something in loader, we can just return empty content

Can you elaborate?

ferdinando-ferreira:
@alexander-akait : from what I understand your suggestion would be
 * don't use the plugin for SSR
 * just pass it to `null-loader` to remove from the final javascript

That would be it?

ferdinando-ferreira:
Closing the PR because from what I've understood it is not implemented correctly and doesn't merit further discussion.

Also opened https://github.com/webpack/webpack/issues/12920 because, with `null-loader` being deprecated, even this won't work in future versions.

alexander-akait:
You don't need `null-loader`, you can do using `resolve`

ferdinando-ferreira:
> you can do using `resolve`

Is it possible to do that without knowing at configuration time the name of the files to be ignored?

alexander-akait:
https://github.com/webpack/webpack/issues/12920 is not solve your problem, CSS modules will not work

alexander-akait:
Let's reopen, after investigate I found what locals can be broken with `null`

ferdinando-ferreira:
Reopened and it works and passes all tests. All that's missing 

> don't need process something in loader, we can just return empty content

right?

alexander-akait (commented):
Let's move it on `loader` level, need just avoid https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/loader.js#L193 and https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/loader.js#L157 when emit false and let's rename option to `emit`

ferdinando-ferreira:
> let's rename option to `emit`

Done. Will move the logic from the plugin to the loader as requested

ferdinando-ferreira:
@alexander-akait : done, it works.

You can check with

```
git clone https://github.com/ferdinando-ferreira/mini-css-extract-plugin --single-branch mini-css-extract-plugin
cd mini-css-extract-plugin
npm install
cd usecase
npm install
npm run build
```
 * `usecase\dist\current-plugin` will have index.css on ssr folder
 * `usecase\dist\intended-behaviour` will **not** have index.css on ssr folder

The only difference in the config is `emit: false`.

It also does not erase `locals` like `null-loader` would.

src/loader.js
```diff
@@ -189,45 +189,49 @@ export function pitch(request) {
...
+    let addDependencies;
+    if (emit) {
+      for (const asset of compilation.getAssets()) {
+        assets[asset.name] = asset.source;
+        assetsInfo.set(asset.name, asset.info);
       }
 
-      const identifierCountMap = new Map();
+      addDependencies = (dependencies) => {
+        if (!Array.isArray(dependencies) && dependencies != null) {
```
ferdinando-ferreira, line 202: This doesn't get covered by test and can be removed. The only caller of this function is a few lines below and is guaranteed to be an array.

Should I remove it?

ferdinando-ferreira:
@alexander-akait : before reviewing, let me create an usecase for `locals` to ensure it works as intended in both cases

ferdinando-ferreira:
@alexander-akait : done! 

To test all usecases and the new one for locals:

```
git clone https://github.com/ferdinando-ferreira/mini-css-extract-plugin --single-branch mini-css-extract-plugin
cd mini-css-extract-plugin
npm install
cd usecase
npm install
npm run build
```

Compilation result will be at `usecase/dist/`.

Then to test the version with locals

```
npm run start:testing-with-locals
```

The result must be
```javascript
{ a: 'foo__style__a', b: 'foo__style__b' }
```

Which is the same result from the example this was copied from: https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/test/cases/commonjs-module-syntax/webpack.config.js

Please review and validate.

ferdinando-ferreira:
@alexander-akait , @cap-Bernardito : I assume #732 is a more elegant way to implement this, thanks for merging.

PR #726, opened 2021-03-20
fix: ensures cssReload is called on first import

### LINKED ISSUE #706: Adding new css import doesn't result in hot style update
- Operating System: macOS BigSur 11.1
- Node Version: `14.2.0`
- NPM Version: `6.14.4`
- webpack Version: `5.23.0`
- mini-css-extract-plugin Version: `1.3.8`

### Expected Behavior

When adding new `.css` import with HMR enabled styles should hot update.

### Actual Behavior

Styles are not hot updated. I get:
```
[HMR] Updated modules:
[HMR]  - ./index.js
[HMR]  - ./second.css
[HMR] App is up to date.
```

But no logs that are usually printed when styles reload - from https://github.com/webpack-contrib/mini-css-extract-plugin/blob/470ba0e553386d7c97e7646169e975322c0e5de2/src/hmr/hotModuleReplacement.js#L230-L236

When I save any `.css` file (can be just resave without any change), only then this newly imported styles are applied:

```
[HMR] Updated modules:
[HMR]  - ./second.css
[HMR] App is up to date.
[HMR] Reload all css
```

### Code

Code in reproduction 👇 

But from what I can tell when new css Module (not css-modules, just webpack module :) ). Is loaded this code gets executed - https://github.com/webpack-contrib/mini-css-extract-plugin/blob/470ba0e553386d7c97e7646169e975322c0e5de2/src/loader.js#L11-L30

But this code only setup `cssReload` function to be executed on next hot update. It doesn't execute it when new module is added for the first time. I guess it might be tricky to potentially execute it only when new module is added, but not on initial code evaluation (first visit in the browser)?

### How Do We Reproduce?

https://github.com/pieh/mini-css-extract-plugin-new-css-import

steps listed in README

ferdinando-ferreira:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Hot module reload was never called on the first inclusion of a css file.

### Breaking Changes

None

### Additional Info
Fixes #706. 

These are the steps to reproduce both the issue and the solution

#### Testing the problem
```
git clone https://github.com/ferdinando-ferreira/mini-css-extract-plugin-new-css-import.git --single-branch mini-css-extract-plugin-new-css-import
cd mini-css-extract-plugin-new-css-import
npm install
```

To test the problem, run

```
npm run start
```

The browser will open on localhost:8080. Then open `index.js´ and remove the comment from the following line

```javascript
// import "./second.css";
```
 * Expected result: font size of foo will increase to 48px
 * Actual result: it doesn't

#### Testing the solution

To test the solution, restore `index.js* to the original state, and run

```
npm run start:fixed
```

The browser will open on localhost:8080. Then open `index.js´ and remove the comment from the following line

```javascript
// import "./second.css";
```
 * Expected result: font size of foo will increase to 48px
 * Actual result: works as expected

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -25,6 +25,7 @@ function hotLoader(content, context) {
   })});
       module.hot.dispose(cssReload);
       ${accept}
+      cssReload();
     }
   `;
 }


ferdinando-ferreira:
Caveat: `cssReload()` is called unnecessarily on first execution.

Considering it has no actual impact that could be a "lesser" bug than the original one (not being called on new import), one that I can fix given enough time and strictly if needed

PR #729, opened 2021-03-23
#691 Auto public path

### LINKED ISSUE #691: Asset links broken when extracting CSS and assets to different folders
- Operating System: OSX
- Node Version: v14.15.4
- NPM Version: 6.14.10
- webpack Version: 5.15.0
- mini-css-extract-plugin Version: ^1.3.5

### Expected Behavior

I'm using `MiniCssExtractPlugin` to extract CSS, and with default settings everything works fine. However I want to export the CSS to its own folder, separate from image assets, like this:

```sh
build/styles/bundle.css
build/assets/image.png
```

Per the [docs](https://webpack.js.org/plugins/mini-css-extract-plugin/), I'm doing this by passing a `filename` option to the plugin with the desired folder name.

### Actual Behavior

When I pass in an option like: `filename: 'styles/[name].css`, the CSS file itself gets exported to the correct directory. However, `url(..)` asset links inside the CSS aren't adjusted, and wind up broken as a result.

Specifically, `styles/bundle.css` winds up containing references like `url(assets/image.png)`, causing the browser to request `styles/assets/image.png`.

### Code

The relevant code snippets look like this:

```js
/* webpack.config.js */
    output: {
        filename: 'js/[name].[contenthash:8].js',
        path: path.resolve('.', 'build'),
        publicPath: '',
    },
    plugins: [
        new VueLoaderPlugin(),
        new HtmlWebpackPlugin({ /* .. */ }),
        new MiniCssExtractPlugin({
            filename: 'styles/[name].css' }
        ),
    ],
    module: {
        rules: [
            // ...
            {
                test: /\.png$/,
                type: 'asset/resource',
                generator: { filename: 'assets/[name][ext]' },
            },
            {
                test: /\.css$/,
                use: [ MiniCssExtractPlugin.loader, 'css-loader' ],
            },
```

```css
/* /src/foo/bar.css */
.whatever {
    background-image: url('~path/to/image.png');
}
```

After running webpack, the CSS winds up with broken asset links:

```css
/* styles/bundle.css */
.whatever {
    background-image: url('assets/image.png');
        /* should be: url('../assets/image.png') or similar */
}
```

### How Do We Reproduce?

By building a project with configuration as described above (in production mode).

-----

Is there some setting or workaround supporting this use case?

Thanks!

karlvr:
This PR contains a:

- [X] **bugfix**
- [X] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Maybe resolves #691 by hopefully correctly handling the "auto" `publicPath` case.

Previously the code set the `publicPath` to `""` when the `publicPath` option was set to `"auto"`. This results in CSS output where all of the assets are relative to the current directory. I believe this would only work if your CSS files were in the root of your output / `publicPath`.

Thew new code computes a relative path from the output file to the root of your output (what `publicPath` represents), and uses that as the effective `publicPath` for all `url(...)`s in the output CSS.

### Breaking Changes

This does change the behaviour of `publicPath = "auto"`. I believe it will still work correctly for everyone that's relying on the existing behaviour actually, so maybe this isn't breaking. I also like that as `"auto"` is the default, it might mean that the output of this plugin works automatically for everyone.

### Additional Info

I have that nagging feeling that because this feels so great, there's something wrong :-) Early days.


karlvr:
I'm not really sure why `chunk.name` would be `null`, so we revert back to the original `"auto"` behaviour in that case and set the effective `publicPath` to `""`.

alexander-akait (commented):
Here logic for this https://github.com/webpack/webpack/blob/982345ecf542a8758dd19f05d5f8a114e37889c6/lib/util/identifier.js#L312

karlvr:
> Here logic for this https://github.com/webpack/webpack/blob/982345ecf542a8758dd19f05d5f8a114e37889c6/lib/util/identifier.js#L312

@alexander-akait thank you. Confirming… that function is what we should use instead of `path.relative(...)`? It doesn't look like it uses filename templates, so it wouldn't replace the `[name]` replacement. Am I on the right page?

alexander-akait:
@karlvr we should look how webpack use it, I don't have time on this right now, maybe on the next week, but you can do it and help me, thanks :smile:

karlvr:
@alexander-akait thanks, it struck me as a fun hack :-D

@andyhall's minimal repo might be a good basis for a test?

I thought I had struggled with tests but I think that must have been another project. All of the tests appear to pass at the moment for me, which is a good start. Code coverage should be resolved with the right tests eh.

alexander-akait:
Fixed by https://github.com/webpack-contrib/mini-css-extract-plugin/pull/772, anyway big thanks for PR and good idea

karlvr:
@alexander-akait this actually doesn't appear to be fixed based on my quick test... using 3.1.1. I have `publicPath: 'auto'` in my `MiniCssExtractPlugin.loader` but it's still outputting relative URLs to assets in the CSS. Am I using it right?

alexander-akait:
Hm, we have many tests on this, can you provide example? Should work…

karlvr:
Gah, sorry I think my build has the issue as well... it's to do with CSS in chunks... it's not getting handled correctly. I'm investigating.

karlvr:
@alexander-akait I'm very sorry, I've found the cause of the problem and it was _my_ version of the library, as my solution to the undo path was to use the file template only if the chunk had a name. No name for chunks created by that `SplitChunksPlugin` sometimes... I thought I'd upgraded back to this library, but I hadn't changed an import. Total false alarm. Now happily back on HEAD. Thank you very much.

PR #737, opened 2021-04-12
perf: add experimental support for `importModule`

sokra:
This replaces the child compiler with a
lightweight execution of modules within
the module graph of the original compilation

To try the new approach.

* Use webpack >= 5.32.0
* `experimentalUseImportModule: true` in `MiniCssPlugin` options
* Avoid `publicPath` in `MiniCssPlugin` options, that doesn't work yet
* Set a static `output.publicPath` in webpack options

This PR contains a:

- [x] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

child compiler take up too much memory and performance

### Breaking Changes

none

### Additional Info

Not all tests are passing yet with the new approach.
There is nothing to do in mini-css-extract-plugin to fix this.
Currently I think that are all webpack problems that I will fix eventually.
Biggest problem currently is that it breaks with HMR.


sokra:
@vyobukhov https://github.com/vyobukhov/mini-css-memory-issues builds within `--max-old-space-size=512` with this PR.

src/loader.js
```diff
@@ -67,6 +187,40 @@ export function pitch(request) {
...
+  if (optionsFromPlugin.experimentalUseImportModule) {
+    if (!this.importModule) {
+      callback(
+        new Error(
+          "You are using experimentalUseImportModule but 'this.importModule' is not available in loader context. You need to have at least webpack 5.33.2."
+        )
+      );
+      return;
+    }
+    console.log(publicPath);
```
alexander-akait, line 199: I think we need avoid it :smile:

alexander-akait (commented):
Need remove `console.log` and I think we can merge and release it

src/loader.js
sokra: Sure. Fixed

sokra:
🚀

alexander-akait (approved):

alexander-akait:
I will do release tomorrow :rocket:

PR #742, opened 2021-04-17
fix: update loader options

chenxsan:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I think the new `experimentalUseImportModule ` is missed from schema.

image

### DIFF
--- a/src/loader-options.json
+++ b/src/loader-options.json
@@ -30,6 +30,9 @@
           "type": "boolean"
         }
       }
+    },
+    "experimentalUseImportModule": {
+      "type": "boolean"
     }
   }
 }


alexander-akait (approved):

alexander-akait:
Oh, this options should be on plugin level

alexander-akait:
We need provide simple case in docs

chenxsan:
You're right, it should be set for plugin instead of loader.

PR #797, opened 2021-08-01
fix(hmr): hmr not work in browser extension

njzydark:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Current `isUrlRequest` in `hotModuleReplacement` is not work in browser extension, because:

Chrome use chrome-extension://
Microsoft Edge use extension://
Firefox use moz-extension://

### DIFF
--- a/src/hmr/hotModuleReplacement.js
+++ b/src/hmr/hotModuleReplacement.js
@@ -198,8 +198,8 @@ function reloadAll() {
 function isUrlRequest(url) {
   // An URL is not an request if
 
-  // It is not http or https
-  if (!/^https?:/i.test(url)) {
+  // It is not http or https or browser extension
+  if (!/^(https?|.*extension):/i.test(url)) {
     return false;
   }
 


alexander-akait:
hmr should work in browser extensions, can you provide example when it will not work

njzydark:
> hmr should work in browser extensions, can you provide example when it will not work

https://github.com/njzydark/chrome-extension-boilerplate/tree/css-hmr-not-work

```
npm i
npm run dev
```

When load extension from chrome and open options page, then change this file https://github.com/njzydark/chrome-extension-boilerplate/blob/css-hmr-not-work/src/options/app.module.less, the css hmr will not work

---

The main branch https://github.com/njzydark/chrome-extension-boilerplate is using my fixed version, css hrm work right

alexander-akait:
Yep, I see, I will fix test tomorrow and do release, thanks

PR #824, opened 2021-09-06
Issue 427

### LINKED ISSUE #427: Support multiple instances of MiniCssExtractPlugin to generate multiple CSS theme output from the single CSS input
- Operating System: Mac OS 10.14.5
- Node Version: v8
- NPM Version: v5
- webpack Version: v4
- mini-css-extract-plugin Version: v0.8.0

### Feature Proposal
Support multiple instances of MiniCssExtractPlugin to generate multiple themed CSS outputs from the shared singular CSS/SCSS input.

<details>
<summary>Sample code</summary>

```js
const __home = path.resolve(__dirname, '');

module.exports = {
    entry: path.resolve('src/index.js'),
    output: {
        filename: "[name].js",
        path: path.resolve("dist")
    },
    module: {
        rules: [
            {
                test: /\.scss$/,
                use: [
                    MiniCssExtractPlugin.loader,
                    "css-loader",
                    {
                        loader: "sass-loader",
                        options: {  includePaths: [path.resolve(__home, "src/light-theme"),]}

                    }
                ]
                // Want to do something like this to generate a second CSS theme bundle based on the shared CSS input extracted from the JS
                // use: [
                //     MiniCssExtractPlugin.loader,
                //     "css-loader",
                //     {
                //         loader: "sass-loader",
                //         options: {  includePaths: [path.resolve(__home, "src/dark-theme"),]}
                //
                //     }
                // ]
            }
        ]
    },
    plugins: [
        new MiniCssExtractPlugin({
            filename: "light-theme.css"
        })
        // new MiniCssExtractPlugin({
        //     filename: "dark-theme.css"
        // })
    ]
};
```

</details>

### Feature Use Case

The use-case for this is an app that wants to support both a dark and light mode styles without having to rely on CSS-in-JS solutions. Instead, it would be nice to extract the CSS from the common singular source CSS and output that into 2 separate stylesheets which use different SCSS variables to generate the dark and light variants of the stylesheet.

This seems feasible if multiple instances of MiniCssExtractPlugin were supported:

One that applies the light-mode theme variables via the configured sass-loader in the first mini-css-extract configuration.

And a second instance that applies the dark-mode theme variables via the configured sass-loader in the second mini-css-extract configuration.

If there is another way to achieve this, I'd be very interested.

## Link to test repo
https://github.com/bjankord/webpack4-multi-theme

alexander-akait:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

fixes #427

### Breaking Changes

No

### Additional Info

Two simple scenrious:
1. loaded default (light) and async loading additional (dark)
2. async loading light and dark

There could be many more scenarios here - you can use `var()` and load only files with with variables, you can use multiple entries and more.

The main idea is using `resourceQuery` and `import(/* webpackChunkName: "dark" */ "./style.scss?dark")` so you can run multiple times `sass-loader`/etc


README.md
```diff
@@ -911,6 +911,150 @@ module.exports = {
...
+  entry: "./src/index.js",
+  module: {
+    rules: [
+      {
+        test: /\.s[ac]ss$/i,
+        oneOf: [
+          {
+            resourceQuery: "?dark",
+            use: [
+              Self.loader,
```
benquarmby, line 931: I don't see a reference to `Self`. Should it be like this?
```suggestion
              MiniCssExtractPlugin.loader,
```
(same for next two occurrences)

README.md
```diff
@@ -911,6 +911,150 @@ module.exports = {
...
+    document.head.appendChild(themes[newTheme]);
+
+    return;
+  }
+
+  if (newTheme === "dark") {
+    // eslint-disable-next-line no-console
+    console.log(`LOADING THEME - ${newTheme}`);
+
+    import(/* webpackChunkName: "dark" */ "./style.scss?dark").then(() => {
```
benquarmby, line 1001: Is the chunk name comment required here, or is it optional? If it's required, would there be  any way to define it in the webpack config rather than the source?

Just thinking out loud: It might be possible to use a pitching loader to expand a static `./style.scss` import to one per theme (i.e. `../style.scss?dark`, `./style.scss?light`, `./style.scss?n...`), and then use matching resource queries.

benquarmby (commented):
Thanks for coming back to theming 💯 
Take comments with a pinch of 🧂 
FYI @greg-a-smith

README.md
alexander-akait: It is for tests

README.md
alexander-akait: This is for tests and examples, for production I recommend to not use comments, because you will have less names for files, so you will send less bytes for client

README.md
benquarmby: Yes, I saw the tests. But this is the README right?

README.md
benquarmby: Got it, thank you. Would it make sense to revise this example to use production best practices?

alexander-akait: Feel free to send a PR and fix it

benquarmby: I can certainly fix the `Self` references in the README. Stand by for that.

However, I don't actually know how to replace `webpackChunkName` comments with something more first class (for example. something we'd see in `webpack.config.js`). This is one of the reasons I've been following this issue so closely.

PR #852, opened 2021-10-09
fix: valid generation `url()` for `auto` value of the `publicPath` op…

### LINKED ISSUE #848: The result of asset/resource's generator.filename will be trimmed unexpectedly
# Bug report

**What is the current behavior?**

When to use `wepbkac@5.52.0` or later, the `asset/resource`'s `generator.filename` result will be trimmed unexpectedly.

**If the current behavior is a bug, please provide the steps to reproduce.**

I don't know it is a bug or intentional. But the behavior was changed from `webpack@5.52.0`.
I've created the reproducible project.

https://github.com/hrsh7th/webpack-asset-resource-repro

**What is the expected behavior?**

The result of CSS's `url(...)` isn't trimmed.

**Other relevant information:**

I think this problem is related to https://github.com/webpack/webpack/issues/14392

---

This issue was moved from webpack/webpack#14424 by @alexander-akait. Original issue was by @hrsh7th.

alexander-akait:
…tion

This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

try to fix #848

### Breaking Changes

No

### Additional Info

Because `auto` allows to do relative paths in `url()`, we just do not allow `new URL(...)` to change URL using `#`, note - you should avoid writing stuff `./foo/../foo/image.png`, because it will be not simplified and will be keeping as it


alexander-akait:
hm, not safe for windows...

PR #899, opened 2022-01-17
Fix cjs

### LINKED ISSUE #896: Breaking change with import via CommonJS in `2.5.0`
### Bug report

Since 2.5.0, importing via CommonJS no longer matches what it did in the past, or what is shown in the documentation.
```js
const MiniCssExtractPlugin = require('mini-css-extract-plugin');
console.log(MiniCssExtractPlugin);
```
```
{
  default: [class MiniCssExtractPlugin] {
    loader: 'E:\\GitHub\\website\\node_modules\\mini-css-extract-plugin\\dist\\loader.js'
  },
  pluginName: 'mini-css-extract-plugin',
  pluginSymbol: Symbol(mini-css-extract-plugin)
}
```

### Actual Behavior
```js
new MiniCssExtractPlugin()
```
This throws a `TypeError: MiniCssExtractPlugin is not a constructor` when attempting to use in the documented way. Using the exported `.default` works fine, but I assume is not the intended behaviour.

### Expected Behavior

It should behave exactly as it did in 2.4.6 and below.  This is likely the result of the change to the `main` export in https://github.com/webpack-contrib/mini-css-extract-plugin/commit/5b5654c9847a615555660d79245b857536f72124#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L14, and the deletion of the `src/cjs.js` script and its build.

Previously in 2.4.6:
```js
const MiniCssExtractPlugin = require('mini-css-extract-plugin');
console.log(MiniCssExtractPlugin);
```
```
[class MiniCssExtractPlugin] {
  loader: 'E:\\GitHub\\website\\node_modules\\mini-css-extract-plugin\\dist\\loader.js'
}
```

alexander-akait:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

fixes #896

### Breaking Changes

No

### Additional Info

Types was changed too, due we use commonjs now, but our real code in commonjs too, so types should use namespaces

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -1,16 +1,16 @@
-import path from "path";
+const path = require("path");
 
-import {
+const {
   findModuleById,
   evalModuleCode,
   AUTO_PUBLIC_PATH,
   ABSOLUTE_PUBLIC_PATH,
   SINGLE_DOT_PATH_SEGMENT,
   stringifyRequest,
-} from "./utils";
-import schema from "./loader-options.json";
+} = require("./utils");
+const schema = require("./loader-options.json");
 
-import MiniCssExtractPlugin, { pluginName, pluginSymbol } from "./index";
+const MiniCssExtractPlugin = require("./index");
 
 /** @typedef {import("schema-utils/declarations/validate").Schema} Schema */
 /** @typedef {import("webpack").Compiler} Compiler */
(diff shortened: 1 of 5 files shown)


src/loader.js
```diff
@@ -478,5 +486,4 @@ export function pitch(request) {
   });
 }
 
-// eslint-disable-next-line func-names
-export default function () {}
+module.exports = { default: function loader() {}, pitch }
```
alexander-akait, line 489: To keep compatibility, ideally we should refactor it in future

PR #900, opened 2022-01-17
fix: [2.5.1] mark getCssModule and getCssDependency as public

barak007:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Mark `getCssModule` and `getCssDependency` as public 
`getCssModule` is the base of our integration to `mini-css-extract-plugin` added support done here #703 
We ended up not using `getCssDependency` but I feel like it should also be public.
These are the basic building blocks for proper integration to css bundling in webpack

### Breaking Changes
Fix types

PR #901, opened 2022-01-17
Fix types

alexander-akait:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

fixes #900

### Breaking Changes

No

### Additional Info

Also fix bug with relative paths for hot code, it was broken before...

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -96,8 +96,8 @@ declare namespace MiniCssExtractPlugin {
   };
 }
 type Compiler = import("webpack").Compiler;
-type CssModuleConstructor = new (...args: any) => any;
-type CssDependencyConstructor = new (...args: any) => CssDependency;
+type CssModuleConstructor = new (...args: TODO) => CssModule;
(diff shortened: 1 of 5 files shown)


alexander-akait:
@barak007 I will merge it and do release, feel free to improve any types (if you found a typo/wrong types/somethings else)

types/index.d.ts
```diff
@@ -187,3 +205,34 @@ type RuntimeOptions = {
   attributes: Record<string, string> | undefined;
 };
 type TODO = any;
+type CssModule = import("webpack").Module & {
+  content: Buffer;
```
barak007, line 209: @alexander-akait can this `content` accept string or Buffer is the only possible value?

alexander-akait: We have `content: Buffer.from(content)` in our code, so I used Buffer, but webpack can accept strings too, need to tests, if it will work feel free to improve it

PR #953, opened 2022-06-15
fix: do not attempt hotreloading when emit is false

### LINKED ISSUE #943: "no window.document found, will not HMR CSS" when emit=false
### Bug report

When using webpack devserver with a server and client bundle where the server bundle

### Actual Behavior

"no window.document found, will not HMR CSS" is output to the console

### Expected Behavior

Since emit = false, there is no need to attempt HMR anyway so simply ignore this

### How Do We Reproduce?

1. checkout https://github.com/ntucker/anansi/tree/25803a5b49316c7e73846efbe774d5cbe3eb28a3
2. yarn install
3. yarn build:pkg
4. cd examples/concurrent
5. yarn start
6. observe logging statement

### Please paste the results of `npx webpack-cli info` here, and mention other relevant information

```
  System:
    OS: Linux 5.10 Ubuntu 20.04 LTS (Focal Fossa)
    CPU: (32) x64 AMD Ryzen 9 5950X 16-Core Processor
    Memory: 26.13 GB / 31.30 GB
  Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    Yarn: 3.2.1 - /mnt/c/Program Files (x86)/Yarn/bin/yarn
    npm: 8.7.0 - ~/.nvm/versions/node/v16.14.2/bin/npm
```

ntucker:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Fixes #943

When doing SSR it is desirable to still run the same css pipeline so consistent hashes can be produced in the SSR code. However, actually generating CSS is unnecessary. Because of this it is custom to set emit to false. In this case, it would be nice to not have `no window.document found, will not HMR CSS` spamming the console as well as same some small computational time.


src/loader.js
```diff
@@ -69,6 +69,7 @@ function hotLoader(content, context) {
 function pitch(request) {
   // @ts-ignore
   const options = this.getOptions(/** @type {Schema} */ (schema));
+  const emit = typeof options.emit !== "undefined" ? options.emit : true;
```
ntucker, line 72: hoisted this so we can access it in the hot reloading path

alexander-akait (approved):

PR #980, opened 2022-10-19
Update README.md to reflect css-modules

justin808:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### DIFF
--- a/README.md
+++ b/README.md
@@ -1017,7 +1017,7 @@ module.exports = {
 
 ### Remove Order Warnings
 
-For projects where css ordering has been mitigated through consistent use of scoping or naming conventions, the css order warnings can be disabled by setting the ignoreOrder flag to true for the plugin.
+For projects where css ordering has been mitigated through consistent use of scoping or naming conventions, such as [CSS Modules](https://github.com/css-modules/css-modules), the css order warnings can be disabled by setting the ignoreOrder flag to true for the plugin.
 
 **webpack.config.js**
 


alexander-akait (approved):
Can you accept CLA?

justin808:
@pulkitkkr and I have been spending hours trying to debug why we get errors on ordering when splitting chunks.

Maybe the code should be updated to never worry about the ordering for CSS-modules automatically?

alexander-akait:
> Maybe the code should be updated to never worry about the ordering for CSS-modules automatically?

That is not safe :disappointed: So we should warning about it

justin808:
What's left to get this merged?

alexander-akait:
Thank you

justin808:
>    Maybe the code should be updated to never worry about the ordering for CSS-modules automatically?
> That is not safe 😞 So we should warning about it

Is that the case where one is mixing some CSS modules and global CSS?

alexander-akait:
I recommend to avoid it, except global styles like fonts/animations/counter styles/etc

justin808:
> I recommend to avoid it, except global styles like fonts/animations/counter styles/etc

What is "it" to avoid? mixing both modules and non-modules?

BTW, we owns https://github.com/shakacode/sass-resources-loader which allows global variables to be used within sass modules.

alexander-akait:
> What is "it" to avoid? mixing both modules and non-modules?

Yes, except global styels for all application, otherwise you need manually to check order in many places

PR #985, opened 2022-10-24
feat(loader): add functions support for locals

yungvldai:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I came across an interesting case in which a custom (modified) css-loader exports functions (as named exports). The [current implementation](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/loader.js#L231) uses `JSON.stringify`, which causes export values of type function ​​to become `undefined`. This small improvement will make it possible to handle function exports.

### Breaking Changes

No braking changes.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -8,6 +8,7 @@ const {
   BASE_URI,
   SINGLE_DOT_PATH_SEGMENT,
   stringifyRequest,
+  stringifyLocal,
 } = require("./utils");
 const schema = require("./loader-options.json");
 
@@ -22,6 +23,7 @@ const MiniCssExtractPlugin = require("./index");
 /** @typedef {import("webpack").AssetInfo} AssetInfo */
 /** @typedef {import("webpack").NormalModule} NormalModule */
 /** @typedef {import("./index.js").LoaderOptions} LoaderOptions */
+/** @typedef {{ [key: string]: string | function } | function} Locals */
 
 /** @typedef {any} TODO */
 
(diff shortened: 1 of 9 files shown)


yungvldai:
Tested locally

alexander-akait (commented):
Can you add test case?

yungvldai:
Wrote unit test. I don't know how to test the integration because I didn't find a css-loader that exports functions in open-source.

alexander-akait (commented):
Can we add an integration test, because unit tests are useless in most of time and don't provide real examples

yungvldai:
Can you provide an example of how can I mock css-loader output? Did not find similar tests in repo.

alexander-akait:
why we need mock css-loader? Also you can create small custom loader to provide example of usage, we need just check work, no need to implement all features in loader, only for test

yungvldai:
Hello! Added tests.

alexander-akait (commented):
Just intresting - do you have a loader with such code generation?

yungvldai:
Yes, I do. I have a fork of css-loader with the support of such code generation, the main idea is to parse BEM entities from classnames and generate ready-to-use functions with all possible mixes & modifiers. I'm currently working on getting this into open source.

yungvldai:
small demo: 
demo

alexander-akait:
Thank you :+1: Sorry for delay

yungvldai:
Thank you too.

PR #1038, opened 2023-06-05
chore: update dependencies to the latest version

ersachin3112:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case
update dependencies to the latest version

### Breaking Changes
No

### Additional Info
No


src/loader.js
```diff
@@ -349,6 +349,7 @@ function pitch(request) {
   const outputOptions = {
     filename: childFilename,
     publicPath,
+    asyncChunkLoading: false,
```
alexander-akait, line 352: We should not do it or fix it on webpack side

ersachin3112: @alexander-akait will https://github.com/webpack/webpack/pull/17365 fix this ?

alexander-akait: @ersachin3112 I think no

src/loader.js
```diff
@@ -373,7 +373,7 @@ function pitch(request) {
   const { NodeTemplatePlugin } = webpack.node;
   const { NodeTargetPlugin } = webpack.node;
 
-  new NodeTemplatePlugin(outputOptions).apply(childCompiler);
+  new NodeTemplatePlugin().apply(childCompiler);
```
ersachin3112, line 376: @alexander-akait  We have only `filename` and `publicPath` in outputOptions but the [NodeTemplatePlugin only uses `asyncChunkLoading`](https://github.com/webpack/webpack/blob/8e0e0eb47d47fe33e4c9ef9bc455d9584f7ea92c/lib/node/NodeTemplatePlugin.js#L32). I think don't need to pass it here, TS also warns about the same.

Screenshot 2023-06-15 at 6 14 48 AM

alexander-akait:
Let's fix old API

PR #1054, opened 2023-10-07
feat: add beforeTagInsert hook

msidolphin:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Additional Info

When I tried webpack-subresource-integrity to inject integrities, I found that I couldn't add integrities to the style files if I also used mini-css-extract-plugin. due to the lack of hooks in mini-css-extract-plugin. Due to the lack of hooks in mini-css-extract-plugin, it is not possible to add additional attributes to the link tag.

#### Related issue

https://github.com/webpack-contrib/mini-css-extract-plugin/issues/40

#### Usage

```js
import MiniCssExtractPlugin from 'mini-css-extract-plugin'

MiniCssExtractPlugin.getHooks(compiler).beforeTagInsert.tap(
  "YourPluginName",
  (source, varNames) =>
    compiler.webpack.Template.asString([
      source,
      `${varNames.tag}.setAttribute("integrity",  __webpack_require__.sriHashes[${varNames.chunkId}]);`,
    ])
);
```


src/hooks.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+function createHooks() {
+  return {
+    beforeTagInsert: new SyncBailHook(["vars"]),
+  };
+}
+
+module.exports = {
+  getHooks,
+};
```
alexander-akait, line 38: Let's make it static methods too

src/index.js
```diff
@@ -912,7 +926,7 @@ class MiniCssExtractPlugin {
                       "}",
                     ]),
                 "return linkTag;",
-              ]
+              ].filter((statement) => statement !== undefined)
```
alexander-akait, line 929: Why?

alexander-akait:
@msidolphin Also can you rebase, sorry for delay

src/index.js
msidolphin: > Why?

Thank you for your reply

Because the following code may return undefined

```js
MiniCssExtractPlugin.getHooks(compiler).beforeTagInsert?.call({
  varNames: {
    chunkId: "chunkId",
    tag: "linkTag",
  },
})
```

if I set this to an empty string, it will cause the test case to time out, and I haven't found a way to update the test case code (other than manually)

```js
MiniCssExtractPlugin.getHooks(compiler).beforeTagInsert?.call({
  varNames: {
    chunkId: "chunkId",
    tag: "linkTag",
  },
}) || "",
```

msidolphin:
> @msidolphin Also can you rebase, sorry for delay

The generated version of package-lock is too old, should I update it?

src/hooks.js
msidolphin: > Let's make it static methods too

getHooks is already a static method in the MiniCssExtractPlugin class, is there anything I need to change here?

alexander-akait:
@msidolphin Just rebase

msidolphin:
> @msidolphin Just rebase

Okay, I'm done.

src/hooks.js
alexander-akait: Let's use this pattern https://github.com/webpack/webpack/blob/main/lib/CleanPlugin.js#L296, also we should use compilation instead compiler

msidolphin: > Let's use this pattern https://github.com/webpack/webpack/blob/main/lib/CleanPlugin.js#L296, also we should use compilation instead compiler

done

alexander-akait (approved):

alexander-akait:
Thank you, I will do release after https://github.com/webpack-contrib/mini-css-extract-plugin/pull/1057

PR #1057, opened 2023-10-24
feat: support named exports with any characters

laverdet:
A proposed change to `css-loader` will allow named exports to be any string, regardless of whether or not it is a valid identifier. The code generation in this project assumes that all exports will be a valid identifier, so the generation stage has been updated to take this into account.

This PR contains a:

- [ ] **bugfix**
- [X] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

When using the new "asIs" named exports feature I submitted in webpack-contrib/css-loader#1549, this module will produce incorrect output since the export names may not be valid local identifiers.

### Breaking Changes
Requires support for `export { local as "string literal" }` syntax which is not supported until nodejs v16.x.

### Additional Info
Named exports resolve the Terser bailout condition described in webpack/webpack#17626 but reduce "grepability" of your code.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -242,23 +242,39 @@ function pitch(request) {
       return;
     }
 
-    const result = locals
-      ? namedExport
-        ? Object.keys(locals)
+    const result = (function makeResult() {
+      // nb: `locals` is reassigned to const to improve TypeScript narrowing.
+      const scopeLocals = locals;
+      if (scopeLocals) {
+        if (namedExport) {
+          const identifiers = Array.from(
+            (function* generateIdentifiers() {
+              let identifierId = 0;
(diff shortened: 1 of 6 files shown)


laverdet:
This is part of a series of pull requests which adds better ecosystem support for exporting original, non-mangled, CSS names as named exports.

webpack-contrib/css-loader#1549
webpack-contrib/mini-css-extract-plugin#1057
seek-oss/css-modules-typescript-loader#50

Due to the optimization bailout condition [webpack/webpack#17626] with JSON default exports consumed by hoistable functions, original non-minified CSS module class names can be found in the minified source code.

PR #1084, opened 2024-02-28
fix: improve esMododule exports

jantimon:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

With `esModule: false` the mini-css-extract-plugin generates JS modules like this:

```js
module.exports = {
  foo: "src-styles-foo",
  bar: "src-styles-bar",
};
```

This allows the following import:

```js
import * as styles from "./styles.css";
import styles from "./styles.css";
import { foo } from "./styles.css";
```

With `esModule: true` the mini-css-extract-plugin generates JS modules like this:

```js
export {
  "foo": "src-styles-foo",
  "bar": "src-styles-bar"
};
```

As there is no default export, the following import will not work:

```js
import styles from "./styles.css"; // Fails
```

Interestingly the test case `es-named-export` tests for it:

https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/test/cases/es-named-export/index.js

```js
import css, { aClass, bClass, cClass } from "./style.css";
```

However when tested manually, the test case fails:

```js
export 'default' (imported as 'css') was not found in './style.css'
```

This PR adds the `defaultExport` option to the mini-css-extract-plugin which will allow the user to write the import exactly as in the es-named-exports test case

### Breaking Changes

This feature is introduces behind an option.
Therefore this is a non-breaking change.
However changing the named export would probably also be not a breaking change)

need to check behaviour https://github.com/webpack/webpack/pull/18271#issuecomment-2032400346


alexander-akait (changes requested):
Let's use `namedExport`, like we have in css-loader

jantimon:
so just kill the option?

jantimon:
@alexander-akait - I added it to the `namedExport` case - is it okay like this?

src/loader.js
```diff
@@ -271,8 +271,10 @@ function pitch(request) {
           const exportsString = `export { ${identifiers
             .map(([id, key]) => `${id} as ${JSON.stringify(key)}`)
             .join(", ")} }`;
-
-          return `${localsString}\n${exportsString}\n`;
+          const exportDefaultString = `export default ${JSON.stringify(
+            locals
+          )}`;
```
sokra, line 276:
```suggestion
          const exportDefaultString = `export default { ${identifiers
            .map(([id, key]) => `${JSON.stringify(key)}: ${id}`)
            .join(", ")} }`;
```

That would be a bit more efficient

jantimon: done:

```diff
--- a/test/cases/es-named-export-as-is/expected/main.js
+++ b/test/cases/es-named-export-as-is/expected/main.js
@@ -17,7 +17,7 @@ var _1 = "Xh041yLR4iCP4RGjge50";
 var _2 = "NMuRsxoDwvW8BhSXhFAY";
 var _3 = "ayWIv09rPsAqE2JznIsI";
 
-/* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = ({"a-class":"Xh041yLR4iCP4RGjge50","b__class":"NMuRsxoDwvW8BhSXhFAY","cClass":"ayWIv09rPsAqE2JznIsI"});
+/* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = ({ "a-class": _1, "b__class": _2, "cClass": _3 });
```

could this somehow prevent the minifier from optimizing?

test/cases/custom-loader-with-functional-exports/expected/main.js
```diff
@@ -8,12 +8,14 @@
 __webpack_require__.r(__webpack_exports__);
 /* harmony export */ __webpack_require__.d(__webpack_exports__, {
 /* harmony export */   cnA: () => (/* binding */ _1),
-/* harmony export */   cnB: () => (/* binding */ _2)
+/* harmony export */   cnB: () => (/* binding */ _2),
+/* harmony export */   "default": () => (__WEBPACK_DEFAULT_EXPORT__)
```
alexander-akait, line 12: We still need an option for this, otherwise we generate extra content, it is bad for a size

alexander-akait: /cc @jantimon

alexander-akait:
@jantimon Just want to clarify, do you want to use and default and named export together?

jantimon:
yes for backwards compatibility - right now nextjs allows both

nextjs uses mini-css-extract-plugin by explicitly setting `esModule: false`

unfortunately `esModule: false` has the drawback that it prevents webpacks esmodule optimizations

so my idea was to switch to esModule but also allow named and default exports just like `esModule: false`

alexander-akait:
@jantimon I see, we need to verify webpack (experimental.css) work the same (or provide an option to do it) and align with this plugin, I will investigate it tomorrow

alexander-akait:
Sorry for delay, I done the same for built-in CSS support - https://github.com/webpack/webpack/pull/18271/ but want some dicussion with trspack team to align behaviour, then it will be merged I will do the same here

alexander-akait:
@jantimon I put it under the options, because we still recommend to use only a named export to be better compatibility with future https://web.dev/articles/css-module-scripts

README.md
```diff
@@ -549,6 +550,60 @@ module.exports = {
...
+Default: `false`
+
+> **Note**
+>
+> This option will work only when you set `namedExport` to `true` in `css-loader`
+
+By default, `mini-css-extract-plugin` generates JS modules based on the `esModule` and `namedExport` options in `css-loader`.
+Using the `esModule` and `namedExport` options will allow you to better optimize your code.
+If you set `esModule: true` and `namedExport: true` for `css-loader` `mini-css-extract-plugin` will generate **only** a named export.
+Our official recommendation is to use only named export for better future compatibility.
```
ahabhgk, line 570: Just for curious, what does "better future compatibility" refer to here? If it's about native CSS modules (https://web.dev/articles/css-module-scripts), it should be distinguishable because of the need for the import-attributes/import-assertions syntax (`assert { type: 'css' }`).

jantimon:
@alexander-akait thanks for merging 🎉

Initially I added the default export also behind a feature flag so it's totally fine for me that you brought that back..

However I have one question - should the following css module code cause a warning for better future compatibility?

**example.module.css**

```css
.default {
  color: red
} 
```

alexander-akait:
@jantimon make sense, but I think better to make it in css-loader

PR #692, opened 2021-02-02
feat: hmr, compare old locals to invalidate if needed

goloveychuk:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
This allows self accept css module if locals didn't changed. If changed - propagate.
Inspired by https://github.com/webpack-contrib/style-loader/blob/master/src/index.js#L231
Now new runtime is 
```js
// extracted by mini-css-extract-plugin
    if(true) {
      // 1612290403401
      var cssReload = __webpack_require__(/*! ../node_modules/mini-css-extract-plugin/dist/hmr/hotModuleReplacement.js */ "../node_modules/mini-css-extract-plugin/dist/hmr/hotModuleReplacement.js")(module.id, {"publicPath":"","esModule":false,"locals":false});
      var _locals = undefined;
      var isEqualLocals = __webpack_require__(/*! ../node_modules/mini-css-extract-plugin/dist/hmr/isEqualLocals.js */ "../node_modules/mini-css-extract-plugin/dist/hmr/isEqualLocals.js");
      module.hot.dispose(function(data) {
        cssReload();
        data.oldLocals = _locals;
      });
      
      if (!_locals || module.hot.invalidate) {
        if (module.hot.invalidate &&
            module.hot.data &&
            module.hot.data.oldLocals &&
            !isEqualLocals(module.hot.data.oldLocals, _locals, false)) {
            module.hot.invalidate();
        } else {
          module.hot.accept();
        }
      }
    }
  //# sourceURL=[module]
//# sourceMappingURL=data:applicati
```

PR #1090, opened 2024-03-20
fix: avoid reloading all csses when hot load

yiminghe:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case
- only refresh page when locals change, otherwise accept and hot load
- avoid reloading all csses 

### Additional Info

Fixes #692


src/loader.js
```diff
@@ -44,9 +44,10 @@ const MiniCssExtractPlugin = require("./index");
  * @returns {string}
  */
 function hotLoader(content, context) {
-  const accept = context.locals
+  const hasLocals = !!(context.locals && Object.keys(context.locals).length);
+  const accept = hasLocals
     ? ""
-    : "module.hot.accept(undefined, cssReload);";
```
alexander-akait, line 49: Why you remove it?

alexander-akait (commented):
We need a test case, please add

src/loader.js
yiminghe: self accept is enough, it will update when dispose: https://webpack.js.org/api/hot-module-replacement/

yiminghe:
> We need a test case, please add

test cases updated

src/loader.js
```diff
@@ -44,22 +44,31 @@ const MiniCssExtractPlugin = require("./index");
...
-    ? ""
-    : "module.hot.accept(undefined, cssReload);";
-
+  const localsJsonString = JSON.stringify(JSON.stringify(context.locals));
   return `${content}
     if(module.hot) {
+      const localsJsonString = ${localsJsonString};
       // ${Date.now()}
-      var cssReload = require(${stringifyRequest(
+      const cssReload = require(${stringifyRequest(
```
alexander-akait, line 52: Don't use const, it will be in the browser and break old browsers

src/loader.js
```diff
@@ -44,22 +44,31 @@ const MiniCssExtractPlugin = require("./index");
  * @returns {string}
  */
 function hotLoader(content, context) {
-  const accept = context.locals
-    ? ""
-    : "module.hot.accept(undefined, cssReload);";
-
+  const localsJsonString = JSON.stringify(JSON.stringify(context.locals));
```
alexander-akait, line 47: Why double `JSON.stringify`? Just typo?

test/cases/hmr/expected/main.js
```diff
@@ -363,10 +363,23 @@ __webpack_require__.r(__webpack_exports__);
...
+        module.hot.data.value !== localsJsonString
+      ) {
+        module.hot.invalidate();
+      } else {
+        module.hot.accept();
+      }
+      module.hot.dispose(function(data) {
+        data.value = localsJsonString;
+        cssReload();
+      });
```
alexander-akait, line 382: We have a special test directory for HMR, will be great to add a test case there, I know your code works fine, but we should ensure about it

alexander-akait (changes requested):
Let's do small improvements and I am fine with it

alexander-akait:
For fixing tests for OLD API just use `OLD_API=1 npm run test:only`

alexander-akait:
Thank you for your PR

src/loader.js
yiminghe: `const localsJsonString = ${localsJsonString};`

string comparation is faster

test/cases/hmr/expected/main.js
yiminghe: what do you mean? I have added cases inside cases folder (hmr, hmr-locals)

yiminghe:
> For fixing tests for OLD API just use `OLD_API=1 npm run test:only`

passed

test/cases/hmr/expected/main.js
alexander-akait: Sorry for delay, just put a test case here https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/test/HMR.test.js

yiminghe: > Sorry for delay, just put a test case here https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/test/HMR.test.js

done

alexander-akait:
@yiminghe We have:
```
var cssReload = require("/home/runner/work/mini-css-extract-plugin/mini-css-extract-plugin/src/hmr/hotModuleReplacement.js")(module.id, undefined);
```

As you can see it is an absolute path, let's use like here https://github.com/webpack-contrib/css-loader/blob/master/src/utils.js#L15

yiminghe:
> @yiminghe We have:
> 
> ```
> var cssReload = require("/home/runner/work/mini-css-extract-plugin/mini-css-extract-plugin/src/hmr/hotModuleReplacement.js")(module.id, undefined);
> ```
> 
> As you can see it is an absolute path, let's use it from here https://github.com/webpack-contrib/css-loader/blob/master/src/utils.js#L15 (maybe we already have the same util here)

updated

alexander-akait (dismissed):

PR #1115, opened 2024-10-25
ci: add Node.js v23

snitin315:
Run tests on Node.js v23


alexander-akait:
Fixed by https://github.com/webpack-contrib/mini-css-extract-plugin/pull/1136

PR #1132, opened 2025-07-18
fix: should contains GetFilenameRuntime for initial chunks

JSerFeng:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Initial css chunk cannot update if user specify `filename` or `chunkFilename`. Because `MiniCssExtractPlugin` has `GetChunkFilenameRuntimeModule` and force the `allChunks` parameter to be `false`, that will cause that runtime cannot get correct url for updated initial css chunk.

This change aligns behavior of webpack's native css support.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1248,7 +1248,7 @@ class MiniCssExtractPlugin {
                 ? this.options.filename
                 : this.options.chunkFilename;
             },
-            false
+            set.has(RuntimeGlobals.hmrDownloadUpdateHandlers)
           )
         );
 


JSerFeng:
How can I add test for this case, seems HMR.test is not what I'm looking for

alexander-akait:
I see, yeah, it is a bug, let's merge, I will fix tests, thanks

PR #1151, opened 2025-12-24
fix(#1041): respect `output.cssFilename` and `output.cssChunkFilename`

### LINKED ISSUE #1041: Use output.cssFilename as default filename
### Feature Proposal

Use `output.cssFilename` as default filename, I know `cssFilename` is used for `experiments.css`, but it would be great if it can also used for this plugin

### Feature Use Case

```js
module.exports = {
  output: {
    cssFilename: "[name]-[contenthash].css",
  },
  plugins: [new MiniCssExtractPlugin()]
}
```

SukkaW:
**Summary**

Fixes #1041 by implementing https://github.com/webpack/mini-css-extract-plugin/issues/1041#issuecomment-3685425730

**What kind of change does this PR introduce?**

Moving `filename` and `chunkFilename` normalization into `apply` (because only then do we have access to the `compiler` object).

Now the precedence is:

```
this.options.filename -> compiler.options.output.cssFilename -> DEFAULT_FILENAME
this.options.chunkFilename -> compiler.options.output.cssChunkFilename -> getChunkFilenameFromFilename(filename)
```

**Did you add tests for your changes?**

Not yet, need help on where to add the test case. I am not familiar with the `test` folder structure here.

**Does this PR introduce a breaking change?**

No, by always preferring passed in `filename` and `chunkFilename` the breaking change could be avoided.

**If relevant, what needs to be documented once your changes are merged or what have you already documented?**

Maybe mention that we will also respect `output.cssFilename` and `output.cssChunkFilename` in existing options docs.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -56,7 +56,7 @@ const {
 
 /**
  * @typedef {object} NormalizedPluginOptions
- * @property {Filename} filename filename
+ * @property {Filename=} filename filename
  * @property {ChunkFilename=} chunkFilename chunk filename
  * @property {boolean} ignoreOrder true when need to ignore order, otherwise false
  * @property {string | ((linkTag: HTMLLinkElement) => void)=} insert a link insert place or a custom insert function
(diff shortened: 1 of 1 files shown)


alexander-akait (approved):

alexander-akait:
Sorry for delay, I will fix CI later

